PR #1, opened 2022-07-12
Application for taking and showing picture initial interface

alexdtat:
Raw interface for android application.


.gitignore
```diff
@@ -0,0 +1,15 @@
+*.iml
+.gradle
+/local.properties
+/.idea/caches
```
TimPushkin: I tend to remove `.idea` folder completely, as we don't rely on it

build.gradle
```diff
@@ -0,0 +1,14 @@
...
+}// Top-level build file where you can add configuration options common to all sub-projects/modules.
+plugins {
+    id 'com.android.application' version '7.2.1' apply false
+    id 'com.android.library' version '7.2.1' apply false
+    id 'org.jetbrains.kotlin.android' version '1.5.31' apply false
+}
+
+task clean(type: Delete) {
+    delete rootProject.buildDir
+}
```
TimPushkin: Let's add a trailing newline in every file that contains code (Gradle, Kotlin, Java) for consistency

app/build.gradle
```diff
@@ -0,0 +1,62 @@
...
+    defaultConfig {
+        applicationId "com.example.featuredetectandroid"
+        minSdk 21
+        targetSdk 32
+        versionCode 1
+        versionName "1.0"
+
+        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+        vectorDrawables {
+            useSupportLibrary true
```
TimPushkin: I don't think we need this as we set `minSdk` to `21`. So, the whole `vectorDrawables` block can be removed

app/build.gradle
```diff
@@ -0,0 +1,62 @@
...
+    }
+    packagingOptions {
+        resources {
+            excludes += '/META-INF/{AL2.0,LGPL2.1}'
+        }
+    }
+}
+
+dependencies {
+
```
TimPushkin: It would look nicer without this newline, in my opinion.

app/build.gradle
```diff
@@ -0,0 +1,62 @@
...
+    implementation "androidx.compose.ui:ui-tooling-preview:$compose_version"
+    implementation 'androidx.lifecycle:lifecycle-runtime-ktx:2.3.1'
+    implementation 'androidx.activity:activity-compose:1.3.1'
+    testImplementation 'junit:junit:4.13.2'
+    androidTestImplementation 'androidx.test.ext:junit:1.1.3'
+    androidTestImplementation 'androidx.test.espresso:espresso-core:3.4.0'
+    androidTestImplementation "androidx.compose.ui:ui-test-junit4:$compose_version"
+    debugImplementation "androidx.compose.ui:ui-tooling:$compose_version"
+    debugImplementation "androidx.compose.ui:ui-test-manifest:$compose_version"
+}
```
TimPushkin: Add a trailing newline, pls

app/build.gradle
```diff
@@ -0,0 +1,62 @@
...
+    packagingOptions {
+        resources {
+            excludes += '/META-INF/{AL2.0,LGPL2.1}'
+        }
+    }
+}
+
+dependencies {
+
+    implementation 'androidx.core:core-ktx:1.7.0'
```
TimPushkin: Android Studio tells me that this (and some other dependencies) can be updated.

Don't update dependencies with `compose_version`, though, as the newer version is `alpha` and may be too unstable.

app/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,29 @@
...
+        android:icon="@mipmap/ic_launcher"
+        android:label="@string/app_name"
+        android:roundIcon="@mipmap/ic_launcher_round"
+        android:supportsRtl="true"
+        android:theme="@style/Theme.FeatureDetectAndroid"
+        tools:targetApi="31">
+        <activity
+            android:name="me.alexdtat.featuredetectandroid.MainActivity"
+            android:exported="true"
+            android:label="@string/app_name"
```
TimPushkin: Android Studio tells me that this is redundant

app/src/main/res/values/colors.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
```
TimPushkin: I don't think you actually need this file when using Jetpack Compose. One color from here is used in `themes.xml`, but its usage can be inlined.

app/src/main/res/values/themes.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+    <style name="Theme.FeatureDetectAndroid" parent="android:Theme.Material.Light.NoActionBar">
+        <item name="android:statusBarColor">@color/purple_700</item>
```
TimPushkin: This is the only place where a color from `colors.xml` is used -- you can inline its value or even remove this line completely as you can control the status bar from Jetpack Compose too. So, the `style` block would look like this:

```xml
<style name="Theme.FeatureDetectAndroid" parent="android:Theme.Material.Light.NoActionBar" />
```

In fact, the whole file can be removed, but there are cases when you need to control app's theme like this, so I would recommend to keep it.

app/src/main/java/me/alexdtat/featuredetectandroid/MainActivity.kt
```diff
@@ -0,0 +1,27 @@
+package me.alexdtat.featuredetectandroid
```
TimPushkin: I think, the name of the main package should correspond to `applicationId` in `gradle.build` and `package` in `AndroidManifest.xml`

TimPushkin (changes requested):
Please, include the `gradle` folder to the files tracked by git. Its absence makes it harder to build the project.

.gitignore
alexdtat: fixed

app/build.gradle
alexdtat: fixed

app/build.gradle
alexdtat: fixed

app/src/main/AndroidManifest.xml
alexdtat: fixed

app/build.gradle
alexdtat: fixed

app/build.gradle
alexdtat: fixed

app/src/main/java/me/alexdtat/featuredetectandroid/MainActivity.kt
alexdtat: Will be `me.alexdtat.featuredetectandroid` ok for the whole app?

app/src/main/res/values/colors.xml
alexdtat: fixed

build.gradle
alexdtat: fixed

app/src/main/res/values/themes.xml
alexdtat: fixed

.gitignore
TimPushkin: It seems like you also accidentally removed `.DS_Store` from `.gitignore`

app/src/main/java/me/alexdtat/featuredetectandroid/MainActivity.kt
TimPushkin: I would actually like something more general, something that will include all three of us, main developers. Let's go with `com.github.featuredetectandroid` as it is pretty common to use the hoster's domain

TimPushkin (changes requested):
Change the package name and fix `.gitignore` as described above in the comments.

.gitignore
alexdtat: Oh yes, actually. Thank you. Fixed.

app/src/main/java/me/alexdtat/featuredetectandroid/MainActivity.kt
alexdtat: I am not very familliar with package naming traditions. Nothing offensive was meant. :)

alexdtat: fixed

TimPushkin (approved):

PR #4, opened 2022-07-17
Getting grayscale image from camera

alexdtat:
Code for getting image from camera, converting it into grayscale image and displaying it on the screen.


alexdtat:
I'm struggling with the architecture. I expect that the code will constantly, in a non-blocking mode, transfer images from the camera to the `ImageAnalysis`, where a grayscale `Plane` will be extracted from it, converted to a `Bitmap` and then displayed. If I understand correctly, I have access to the image only from the `analysis` function. However: 1) it's not `@Composable`, so `@Composable` functions can't be called inside it, 2) I don't like that the analysing function also displays some content, but I haven't figured out how to avoid it.

P.S. I didn't finish and test  this code because I couldn't compile it, so it's not the final version.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -24,4 +55,100 @@ class MainActivity : ComponentActivity() {
...
+            get(data)
+            return data
+        }
+
+        override fun analyze(image: ImageProxy) {
+            width = image.width
+            height = image.height
+            val buffer = image.planes[0].buffer
+            grayScaleByteArray = buffer.toByteArray()
+            ViewGrayscale(grayScaleByteArray, width, height)
```
TimPushkin: I didn't look at the code much as it is WIP and, as I see, the problem lies here.

You can, for example, create a `ViewModel` with `MutableState` fields representing the arguments that are passed here. When created, `ViewGrayscale` will take this `ViewModel` as argument and work with its fields.

TimPushkin: So, here you'll just set these fields to the values you need

alexdtat:
I hardcoded the 90 deg right rotation because of wrong orientation of images got from camera.

alexdtat:
Current result of app's image proccessing.

image

app/src/main/java/com/github/featuredetectandroid/utils/GrayscaleByteArrayToBitmap.kt
```diff
@@ -0,0 +1,32 @@
...
+private const val Y_SHIFT = 127 // Y-component should be inverted and shifted
+
+private const val MAX_COLOR = 0xff
+private const val MAX_ALPHA = MAX_COLOR shl ALPHA_SHIFT
+
+fun grayscaleByteArrayToBitmap(grayscaleByteArray: ByteArray, width: Int, height: Int): Bitmap {
+    val byteList = grayscaleByteArray.toList()
+    Log.i(TAG, "Converting ByteArray image of size $width x $height to a Bitmap.")
+    val buffer = IntBuffer.allocate(grayscaleByteArray.size).apply {
+        byteList.forEach { byte ->
```
TimPushkin: Here you can do `for (byte in grayscaleByteArray)` and then you won't need to convert it to `List`.

Usually `for (x in y)` is preferable over `y. forEach { x -> ... }` (when possible) anyway as it is a reserved word made for this specifically making the code more readable.

app/src/main/java/com/github/featuredetectandroid/utils/GrayscaleByteArrayToBitmap.kt
```diff
@@ -0,0 +1,32 @@
...
+import android.util.Log
+import java.nio.IntBuffer
+
+private const val TAG = "Conversions"
+
+private const val ALPHA_SHIFT = 24 // Alpha channel shift
+private const val RED_SHIFT = 16 // Red channel shift
+private const val GREEN_SHIFT = 8 // Green channel shift
+private const val BLUE_SHIFT = 0 // Blue channel shift
+private const val Y_SHIFT = 127 // Y-component should be inverted and shifted
```
TimPushkin: It's better to separate this shift from the other (by one empty line) because it's a little different from them

app/src/main/java/com/github/featuredetectandroid/ui/GrayscaleViewModel.kt
```diff
@@ -0,0 +1,40 @@
...
+import com.github.featuredetectandroid.utils.grayscaleByteArrayToBitmap
+
+class GrayscaleViewModel : ViewModel() {
+    private var grayscaleByteArray by mutableStateOf(byteArrayOf())
+    private var width by mutableStateOf(0)
+    private var height by mutableStateOf(0)
+
+    @Suppress("FunctionNaming")
+    @Composable
+    fun ViewGrayscale() {
```
TimPushkin: `ViewModel` is not a place for composable (or any other UI elements that don't need their state to be holded): you can read more about it [here](https://developer.android.com/topic/architecture#ui-layer).

It is a good practice to create separate files for composables as you do for classes.

app/src/main/java/com/github/featuredetectandroid/ui/GrayscaleViewModel.kt
```diff
@@ -0,0 +1,40 @@
...
+import androidx.compose.ui.graphics.asImageBitmap
+import androidx.lifecycle.ViewModel
+import com.github.featuredetectandroid.utils.grayscaleByteArrayToBitmap
+
+class GrayscaleViewModel : ViewModel() {
+    private var grayscaleByteArray by mutableStateOf(byteArrayOf())
+    private var width by mutableStateOf(0)
+    private var height by mutableStateOf(0)
+
+    @Suppress("FunctionNaming")
```
TimPushkin: Is this required for some code analysis tool? Android Studio seems to have no problems with the special composables naming.

app/src/main/java/com/github/featuredetectandroid/ui/GrayscaleViewModel.kt
TimPushkin: If you think there will be a problem with reading `GrayscaleViewModel`'s fields, I think, there shouldn't be any if you pass them as arguments (you can even pass the whole `ViewModel` but it's not needed here) because they are `MutabelState`s.

app/src/main/java/com/github/featuredetectandroid/ui/GrayscaleViewModel.kt
```diff
@@ -0,0 +1,40 @@
...
+    @Composable
+    fun ViewGrayscale() {
+        if (width == 0 || height == 0) return
+        val grayscaleBitmap = grayscaleByteArrayToBitmap(
```
TimPushkin: I don't think a composable should do such calculations: composables should be as light as possible. Why don't you place it into `setPicture` and just store the resulting `Bitmap` in the `ViewModel`?

app/src/main/java/com/github/featuredetectandroid/ui/GrayscaleViewModel.kt
```diff
@@ -0,0 +1,40 @@
...
+            grayscaleByteArray,
+            width,
+            height
+        )
+        Image(
+            bitmap = grayscaleBitmap.asImageBitmap(),
+            modifier = Modifier.fillMaxSize(),
+            contentDescription = "Grayscale photo"
```
TimPushkin: Just for your knowledge: in production you are required to localize `contentDescription`s as they are used for [accessibility purposes](https://developer.android.com/jetpack/compose/accessibility#describe-visual).

Leave this as is here, though, because we want to keep the UI code as simple as possible for now.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,153 @@
...
+    private val imageViewModel by viewModels<GrayscaleViewModel>()
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+
+        checkPermissions()
+        cameraExecutor = Executors.newSingleThreadExecutor()
         setContent {
-            Theme {
+            MaterialTheme {
```
TimPushkin: This should be the theme you defined in `Theme.kt` (so, `Theme` here)

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,153 @@
...
 
 class MainActivity : ComponentActivity() {
+    private lateinit var cameraExecutor: ExecutorService
+    private val imageViewModel by viewModels<GrayscaleViewModel>()
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+
+        checkPermissions()
+        cameraExecutor = Executors.newSingleThreadExecutor()
```
TimPushkin: Why not place the initialization  where the property is declared without making it `lateinit`? It's not something that needs `Activity` to be initialized.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,153 @@
...
+    private val cameraPermission =
+        registerForActivityResult(ActivityResultContracts.RequestPermission()) { granted ->
+            when {
+                granted -> {
+                    Toast.makeText(
+                        this,
+                        "Permissions granted.",
+                        Toast.LENGTH_SHORT
+                    ).show()
+                    finish()
```
TimPushkin: When I start the app the first time and grant the permissions this makes the app close and I have to reopen it manually again

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,153 @@
...
+                    Log.d(TAG, "Permission was granted successfully.")
+                    startCamera()
+                } else -> {
+                    Toast.makeText(
+                        this,
+                        "Permissions not granted by the user.",
+                        Toast.LENGTH_SHORT
+                    ).show()
+                    finish()
```
TimPushkin: I think a better solution would be to still start the app but display something like "Camera permission required" in place of the image

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,153 @@
...
+    }
+
+    override fun onDestroy() {
+        super.onDestroy()
+        cameraExecutor.shutdown()
+    }
+
+    private val cameraPermission =
+        registerForActivityResult(ActivityResultContracts.RequestPermission()) { granted ->
+            when {
```
TimPushkin: `if`-`else` would be simpler here

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,153 @@
...
+            startCamera()
+        } else {
+            cameraPermission.launch(Manifest.permission.CAMERA)
+        }
+    }
+    private fun allPermissionsGranted() = REQUIRED_PERMISSIONS.all {
+        ContextCompat.checkSelfPermission(
+            this,
+            it
+        ) == PackageManager.PERMISSION_GRANTED
```
TimPushkin: When `checkSelfPermissions` returns `DENIED` you should call `shouldShowRequestPermissionRationale` for this permission and then probably explain to a user why you need this permission. The workflow is nicely described [here](https://developer.android.com/training/permissions/requesting#request-permission) with a code snippet.

TimPushkin: Maybe you'll be able to simplify your code as you only need one permission

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,153 @@
...
+                        Toast.LENGTH_SHORT
+                    ).show()
+                    finish()
                 }
             }
         }
+
+    private fun checkPermissions() {
+        if (allPermissionsGranted()) {
+            startCamera()
```
TimPushkin: It's strange that a function called `checkPermissions` starts the camera don't you think?

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,153 @@
...
+    }
+
+    private fun startCamera() {
+        val cameraProviderFuture = ProcessCameraProvider.getInstance(this)
+
+        cameraProviderFuture.addListener({
+            val cameraProvider: ProcessCameraProvider = cameraProviderFuture.get()
+
+            val imageAnalyzer = ImageAnalysis.Builder()
+                .setOutputImageFormat(OUTPUT_IMAGE_FORMAT_YUV_420_888)
```
TimPushkin, line 117: [Docs say](https://developer.android.com/reference/androidx/camera/core/ImageAnalysis.Builder#setTargetRotation(int)) we can also set image rotation here. Won't it help with the rotation problem you have?

TimPushkin, line 117: You probably also want to set [an operation mode](https://developer.android.com/training/camerax/analyze#operating-modes) here. Remember we wanted to use non-blocking?

TimPushkin: There is also an example of getting image rotation in the [code snippet here](https://developer.android.com/training/camerax/analyze#bind-to-lifecycle)

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,153 @@
...
+                .build()
+                .also {
+                    it.setAnalyzer(cameraExecutor, PhotoAnalyzer(imageViewModel))
```
TimPushkin: Won't `apply` be better here? This `it` will be eliminated this way.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,153 @@
...
+                )
+            } catch (illegalState: IllegalStateException) {
+                Log.e(TAG, "Use case binding failed: ", illegalState)
+            } catch (illegalArgument: IllegalArgumentException) {
+                Log.e(TAG, "Use case binding failed: ", illegalArgument)
+            }
+        }, ContextCompat.getMainExecutor(this))
+    }
+
+    private class PhotoAnalyzer(val imageViewModel: GrayscaleViewModel) : ImageAnalysis.Analyzer {
```
TimPushkin: Maybe extract this class to another file?

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,153 @@
...
+        private fun ByteBuffer.toByteArray(): ByteArray {
+            rewind()
+            val data = ByteArray(remaining())
+            get(data)
+            return data
```
TimPushkin: I would write this three lines as `return ByteArray(remaining()).also { get(it) }` but maybe I love scope functions too much...

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,153 @@
...
+        }
+
+        private fun rotateRightOn90Deg(
+            grayscaleByteArray: ByteArray,
+            width: Int,
+            height: Int
+        ) = ByteArray(width * height).apply {
+            for (i in 0 until width) {
+                for (j in 0 until height) {
+                    this[i * height + j] = grayscaleByteArray[i + (height - 1 - j) * width]
```
TimPushkin: Replacing `apply` with `also { array -> ...` and `this` with `array` seems better to me

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,153 @@
...
+                }
+            }
+        }
+
+        // Here was hardcoded the 90 deg right rotation because of wrong orientation of images got from camera.
```
TimPushkin: It should be "the 90 deg right rotation is hardcoded".

And isn't this line too long?

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,153 @@
...
+            val height = image.height
+            val buffer = image.planes[0].buffer
+            val grayScaleByteArray = buffer.toByteArray()
+            val rotated = rotateRightOn90Deg(grayScaleByteArray, width, height)
+            imageViewModel.setPicture(rotated, height, width)
+            image.close()
+        }
+    }
+
+    companion object {
```
TimPushkin: Why have you chosen `companion object` over top-level constants, what are the pros of this? I see the following cons:
1. This creates a separate object associated with the class.
2. This requires more lines of code.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,153 @@
...
+        private const val TAG = "FeatureDetectApp"
```
TimPushkin: `TAG` is usually the name of the class. It's just a convention but I think debugging is much easier this way -- you immediately see where to look from the logs.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,153 @@
...
+        private val REQUIRED_PERMISSIONS = mutableListOf(Manifest.permission.CAMERA).toTypedArray()
```
TimPushkin: What do you need `toTypedArray()` for?

app/src/main/AndroidManifest.xml
```diff
@@ -2,6 +2,12 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
     package="com.github.featuredetectandroid">
+    <uses-feature
+        android:name="android.hardware.camera2"
+        android:required="true"/>
+    <uses-permission android:name="android.permission.CAMERA" />
+    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"
```
TimPushkin: I didn't see this used anywhere

app/build.gradle
```diff
@@ -40,9 +40,17 @@ android {
             excludes += '/META-INF/{AL2.0,LGPL2.1}'
         }
     }
+    buildFeatures {
+        viewBinding true
```
TimPushkin: Maybe I missed something, but I didn't see any view binding in the code

TimPushkin (changes requested):
Seems great overall!

app/src/main/java/com/github/featuredetectandroid/ui/GrayscaleViewModel.kt
alexdtat: > Is this required for some code analysis tool? Android Studio seems to have no problems with the special composables naming.

It required for detekt. I planned to add it to the CI, but I'm not sure how useful it will be.

app/build.gradle
alexdtat: fixed

app/src/main/AndroidManifest.xml
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/utils/GrayscaleByteArrayToBitmap.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/utils/GrayscaleByteArrayToBitmap.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
UPD: Using `when` because of `shouldShowRequestPermissionRationale`.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
alexdtat: In detekt max line size is aroud 120 symbols IIRC. Though I can split it.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
alexdtat: But the documentation says that non-blocking mode is set by default. However, I can do it explicitly.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
TimPushkin: Oh, ok than, no need to.

app/src/main/java/com/github/featuredetectandroid/ui/GrayscaleViewModel.kt
TimPushkin: It'll be useful in case you forget to run it or somebody else (me, for example) commits

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
TimPushkin: 100 is the default in Android Studio, let's keep it this way

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
alexdtat: Rudimentary code. Fixed.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
alexdtat: Will be `tryStartCamera` suitable function name? Or `try` is too strongly associated with exceptions?

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
alexdtat: > [Docs say](https://developer.android.com/reference/androidx/camera/core/ImageAnalysis.Builder#setTargetRotation(int)) we can also set image rotation here. Won't it help with the rotation problem you have?

Unfortunately, it seems, this method does nothing in our case (I've tried all possible rotations with the same result). But maybe I missed something.

UPD: A little misunderstanding. Yes, we can use it to improve the hardcoded rotation.

alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
alexdtat: In the past I was advised to use `companion object` for constants. Though I googled a little bit and it seems to be a not welcomed practise.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/ui/GrayscaleViewModel.kt
alexdtat: > I don't think a composable should do such calculations: composables should be as light as possible. Why don't you place it into `setPicture` and just store the resulting `Bitmap` in the `ViewModel`?

`Bitmap` needs `width` > 0 and `height` > 0 for construction so `mutableStateOf` will have at least 1x1 `Bitmap` image as argument.
UPD: I use nullable varibale. Fixed.

app/src/main/java/com/github/featuredetectandroid/ui/GrayscaleViewModel.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/utils/GrayscaleByteArrayToBitmap.kt
TimPushkin: You don't need `toList` conversion now, so remove it

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
TimPushkin: It's ok

app/build.gradle
```diff
@@ -43,6 +43,11 @@ android {
 }
 
 dependencies {
+    def camerax_version = "1.1.0-beta01"
```
TimPushkin: You don't need to use `"..."` here as it's not a formatted string -- `'...'` would be enough

TimPushkin: There are several other places in this file where you can replace double quotes with single ones. Not a big deal, but still.

app/build.gradle
```diff
@@ -43,6 +43,11 @@ android {
...
+    implementation "androidx.camera:camera-core:${camerax_version}"
+    implementation "androidx.camera:camera-camera2:${camerax_version}"
+    implementation "androidx.camera:camera-lifecycle:${camerax_version}"
```
TimPushkin: You don't need to use braces in any of these lines

app/src/main/AndroidManifest.xml
```diff
@@ -2,6 +2,10 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
     package="com.github.featuredetectandroid">
+    <uses-feature
```
TimPushkin, line 6: If you press `Ctrl+Alt+L` in this file, Android Studio will fix some formatting issues

app/src/main/java/com/github/featuredetectandroid/ui/GrayscaleView.kt
```diff
@@ -0,0 +1,23 @@
...
+@Composable
+@Suppress("FunctionNaming")
+fun GrayscaleView(
+    grayscaleBitmap: Bitmap?,
+    width: Int,
+    height: Int
+) {
+    if (width == 0 || height == 0) return
+    Image(
+        bitmap = grayscaleBitmap?.asImageBitmap() ?: return,
```
TimPushkin: I think, you can just do `grayscaleBitmap?.let { bitmap -> ...` and remove all other checks. If somebody passes image of zero size it's their fault.

TimPushkin: Maybe you don't even need this composable because it is just an `Image` and the `ViewModel` because it only stores the bitmap? Such `ViewModel` could be replaced with `rememberSaveable`, though, it may have problems saving large bitmaps, so I'm not sure.

app/src/main/java/com/github/featuredetectandroid/ui/theme/Theme.kt
```diff
@@ -21,6 +21,7 @@ private val LightColorPalette = lightColors(
 )
 
 @Composable
+@Suppress("FunctionNaming")
 fun Theme(darkTheme: Boolean = isSystemInDarkTheme(), content: @Composable () -> Unit) {
```
TimPushkin: Maybe rename this to something more descriptive? Now it looks like something built-in in Jetpack Compose.

app/src/main/java/com/github/featuredetectandroid/utils/GrayscaleArrayToBitmap.kt
```diff
@@ -0,0 +1,36 @@
...
+private const val RED_SHIFT = 16 // Red channel shift
+private const val GREEN_SHIFT = 8 // Green channel shift
+private const val BLUE_SHIFT = 0 // Blue channel shift
+
+private const val Y_SHIFT = 127 // Y-component should be inverted and shifted
+
+private const val MAX_COLOR = 0xff
+private const val MAX_ALPHA = MAX_COLOR shl ALPHA_SHIFT
+
+fun grayscaleArrayToBitmap(grayscaleByteArray: ByteArray, width: Int, height: Int): Bitmap {
```
TimPushkin: Y in YUV is called "luminance", so, as you are expecting it here, it's probably better to replace "grayscale" with "luminance"

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt
```diff
@@ -0,0 +1,50 @@
...
+private const val TAG = "PhotoAnalyzer"
+private const val ROTATION_ON_90_DEGREES = 90
+
+class PhotoAnalyzer(private val imageViewModel: GrayscaleViewModel) : ImageAnalysis.Analyzer {
+    private fun ByteBuffer.toByteArray(): ByteArray {
+        rewind()
+        return ByteArray(remaining()).also { get(it) }
+    }
+
+    private fun rotateRightOn90Deg(
```
TimPushkin: "Right" is not very descriptive when we are talking about rotations. It should be either "clockwise" or "counterclockwise".

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt
```diff
@@ -0,0 +1,50 @@
...
+    because of wrong orientation of some images got from camera. */
+    override fun analyze(image: ImageProxy) {
+        val rotationDegrees = image.imageInfo.rotationDegrees
+        Log.i(TAG, "The image is rotated on $rotationDegrees degrees.")
+
+        var width = image.width
+        var height = image.height
+        val buffer = image.planes[0].buffer
+        val grayScaleByteArray = buffer.toByteArray()
+        val oriented = if (rotationDegrees == ROTATION_ON_90_DEGREES) {
```
TimPushkin: `rotationDegrees`'s docs say that it can be any of 0, 90, 180, 270 degrees. As a temporary solution you can apply the 90-degree rotation `rotationDegrees / 90` times and add a `TODO` to fix this later.

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt
```diff
@@ -0,0 +1,50 @@
+package com.github.featuredetectandroid.utils
+
+import android.util.Log
+import androidx.camera.core.ImageAnalysis
+import androidx.camera.core.ImageProxy
+import com.github.featuredetectandroid.ui.GrayscaleViewModel
+import java.nio.ByteBuffer
+
+private const val TAG = "PhotoAnalyzer"
+private const val ROTATION_ON_90_DEGREES = 90
```
TimPushkin: Something like `ROTATION_STEP` would be better

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt
```diff
@@ -0,0 +1,50 @@
...
+    ) = ByteArray(width * height).also { array ->
+        for (i in 0 until width) {
+            for (j in 0 until height) {
+                array[i * height + j] = grayscaleByteArray[i + (height - 1 - j) * width]
+            }
+        }
+    }
+
+    /* Here is the 0 or 90 deg (depending on state) right rotation hardcoded
+    because of wrong orientation of some images got from camera. */
```
TimPushkin: Sounds not really good in English for me, though, I'm not a native speaker too. It would be better to make this comment shorter and place this where the rotation happens.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,138 @@
...
             Theme {
                 Box(Modifier.fillMaxSize()) {
-                    FloatingActionButton(
-                        onClick = { /*TODO*/ },
-                        modifier = Modifier.align(Alignment.BottomEnd)
-                    ) { }
+                    if (ContextCompat.checkSelfPermission(
+                            this@MainActivity,
+                            Manifest.permission.CAMERA
+                        ) == PackageManager.PERMISSION_GRANTED
```
TimPushkin: It seems like this check is run only when the app starts and isn't rerun after the permission is granted. So, the following happens:
- I start the app the first time
- It asks me for permissions and this check fails
- I grant the permissions
- Camera view doesn't appear because the check is not rerun

And I need to restart the app to see the output :)

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,138 @@
...
+                    ) {
+                        GrayscaleView(
+                            imageViewModel.grayscaleBitmap,
+                            imageViewModel.width,
+                            imageViewModel.height
+                        )
+                    } else {
+                        Text(text = "Camera permission required")
```
TimPushkin: This text appears in the corner of the screen. You need to configure center alignment in the `Box`.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,138 @@
...
+                    Toast.LENGTH_SHORT
+                ).show()
+            }
+        }
+
+    private fun tryStartCamera() {
+        when {
+            cameraPermissionGranted() -> {
+                startCamera()
+            }
```
TimPushkin: Braces are not needed here

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,138 @@
...
+            rationaleSDKVersionCheck() &&
+                shouldShowRequestPermissionRationale(Manifest.permission.CAMERA) -> {
```
TimPushkin: Maybe you can put these two calls into a single function? It'll look better.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,138 @@
...
+                Toast.makeText(
+                    this,
+                    "Without camera permission app can't get and display keypoints.",
+                    Toast.LENGTH_SHORT
+                ).show()
+            }
+            else -> {
+                cameraPermission.launch(Manifest.permission.CAMERA)
+            }
```
TimPushkin: Braces are not needed here too

TimPushkin (changes requested):
More changes required

app/src/main/java/com/github/featuredetectandroid/ui/theme/Theme.kt
alexdtat: > Maybe rename this to something more descriptive? Now it looks like something built-in in Jetpack Compose.

Yes, even I mixed them up last time...

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
alexdtat: > Braces are not needed here

Wouldn't it be better to use braces here for reasons of uniformity?

TimPushkin: I think, it's more readable without additional braces, because it takes less space. Style guide also seems to [permit this](https://developer.android.com/kotlin/style-guide#braces).

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/AndroidManifest.xml
alexdtat: fixed

app/build.gradle
alexdtat: fixed

app/build.gradle
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/ui/theme/Theme.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/utils/GrayscaleByteArrayToBitmap.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt

app/src/main/java/com/github/featuredetectandroid/utils/GrayscaleArrayToBitmap.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt

app/src/main/java/com/github/featuredetectandroid/ui/GrayscaleView.kt
alexdtat: I inlined `View` and kept `ViewModel`.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
alexdtat: Fixed, if I understood you correctrly.

alexdtat:
https://github.com/TimPushkin/FeatureDetectAndroid/pull/4#discussion_r925884182

The question of using `rememberSaveable` instead of `ViewModel` is still open.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
TimPushkin: Nope, this still happens. To reproduce the behaviour do the following:
1. Remove the app's camera permission (or just reinstall it).
2. Start the app.
3. Grant camera permissions.
4. Still see the "Camera permission required" text.

I think, you need to create a `Boolean` field in your `ViewModel` and track permission status there so that Compose sees its updates.

TimPushkin:
> The question of using `rememberSaveable` instead of `ViewModel` is still open.

Yeah, seems like you need to keep this `ViewModel`.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,146 @@
...
         super.onCreate(savedInstanceState)
+
+        cameraPermission =
+            registerForActivityResult(ActivityResultContracts.RequestPermission()) { granted ->
+                if (granted) {
+                    Toast.makeText(
+                        this,
+                        "Permissions granted.",
+                        Toast.LENGTH_SHORT
+                    ).show()
```
TimPushkin: It can be squeezed into one line

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,146 @@
...
+                    Log.d(TAG, "Permission was granted successfully.")
+                    startCamera()
+                } else {
+                    Toast.makeText(
+                        this,
+                        "Permissions not granted by the user.",
+                        Toast.LENGTH_SHORT
+                    ).show()
```
TimPushkin: You can replace "Permissions not granted by the user." with "Permissions not granted." for uniformity and also squeeze it into a single line

TimPushkin: Why do we log that permissions were granted but don't log that they weren't?

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,146 @@
...
-                        onClick = { /*TODO*/ },
-                        modifier = Modifier.align(Alignment.BottomEnd)
-                    ) { }
+                    if (ContextCompat.checkSelfPermission(
+                            this@MainActivity,
+                            Manifest.permission.CAMERA
+                        ) == PackageManager.PERMISSION_GRANTED
+                    ) {
+                        Image(
+                            bitmap = imageViewModel.grayscaleBitmap?.asImageBitmap() ?: return@Box,
```
TimPushkin: Labeled statements should be avoided when possible because they are hard to understand. Just surround the `Image` with `imageViewModel.grayscaleBitmap?.let { bitmap -> ... }` -- this would look much less cumbersome.

TimPushkin: I thought detekt has warnings about labeled statements

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,146 @@
...
+    private fun cameraPermissionGranted() = ContextCompat.checkSelfPermission(
+        this,
+        Manifest.permission.CAMERA
+    ) == PackageManager.PERMISSION_GRANTED
+
+    private fun rationaleSDKVersionCheck() =
+        android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.M
+
+    private fun tryShowRequestPermissionRationale() = rationaleSDKVersionCheck() &&
+        shouldShowRequestPermissionRationale(Manifest.permission.CAMERA)
```
TimPushkin: Unite `rationaleSDKVersionCheck()` and `tryShowRequestPermissionRationale()` into a single function and name it something like `shouldRationalize()`

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,146 @@
...
+
+            val cameraSelector = CameraSelector.DEFAULT_BACK_CAMERA
+
+            try {
+                cameraProvider.unbindAll()
+                cameraProvider.bindToLifecycle(
+                    this,
+                    cameraSelector,
+                    imageAnalyzer
+                )
```
TimPushkin: This can fit into a single line. Personally I only see a point in writing arguments like this in two cases:
1. They don't fit into a single line.
2. Several named arguments are used.

app/src/main/java/com/github/featuredetectandroid/ui/GrayscaleViewModel.kt
```diff
@@ -0,0 +1,20 @@
...
+
+class GrayscaleViewModel : ViewModel() {
+    var grayscaleBitmap: Bitmap? by mutableStateOf(null)
+
+    fun setPicture(grayscaleByteArray: ByteArray, newWidth: Int, newHeight: Int) {
+        grayscaleBitmap = luminanceArrayToBitmap(
+            grayscaleByteArray,
+            newWidth,
+            newHeight
+        )
```
TimPushkin: This can fit into a single line

app/src/main/java/com/github/featuredetectandroid/ui/theme/FeatureDetectAppTheme.kt
```diff
@@ -21,7 +21,11 @@ private val LightColorPalette = lightColors(
 )
 
 @Composable
-fun Theme(darkTheme: Boolean = isSystemInDarkTheme(), content: @Composable () -> Unit) {
+@Suppress("FunctionNaming")
```
TimPushkin: I don't have much experience with detekt but we can probably configure it as described [here](https://detekt.dev/docs/introduction/compose/#functionnaming) to remove such supressions

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt
```diff
@@ -0,0 +1,50 @@
...
+    ) = ByteArray(width * height).also { array ->
+        for (i in 0 until width) {
+            for (j in 0 until height) {
+                array[i * height + j] = grayscaleByteArray[i + (height - 1 - j) * width]
+            }
+        }
+    }
+
+    // Rotation of incorrectly oriented images is implemented here.
+    @Suppress("ForbiddenComment")
```
TimPushkin: Maybe it would be better to put this comment where the rotation happens (inside the function) and get rid of this supression?

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt
```diff
@@ -0,0 +1,50 @@
...
+    override fun analyze(image: ImageProxy) {
+        val rotationDegrees = image.imageInfo.rotationDegrees
+        Log.i(TAG, "The image is rotated on $rotationDegrees degrees.")
+
+        var width = image.width
+        var height = image.height
+        var oriented = image.planes[0].buffer.toByteArray()
+
+        // TODO: Optimize rotations for 180 and 270 degrees.
+        repeat((0 until (rotationDegrees / ROTATION_STEP)).count()) {
```
TimPushkin: Isn't it the same as `repeat(rotationDegrees / ROTATION_STEP) { ... }`?

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt
```diff
@@ -0,0 +1,50 @@
...
+            oriented = rotateClockwiseOnRotationStep(oriented, width, height).also {
```
TimPushkin: Why do you need `also` here?

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt
```diff
@@ -0,0 +1,50 @@
...
+                width = height.also { height = width }
+            }
+        }
+
+        imageViewModel.setPicture(oriented, width, height)
```
TimPushkin, line 45: I'm a bit frustrated, probably you can explain this to me. So, we need to do two transformations: rotation and bitmap conversion. Why one of them happens inside `PhotoAnalyzer` and the other one in `ViewModel`?

TimPushkin (changes requested):
We're getting close 🙃

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt
alexdtat: > Maybe it would be better to put this comment where the rotation happens (inside the function) and get rid of this supression?

"ForbiddenComment" is comment with "TODO".

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
alexdtat: No, as far as I know from my experience, detekt does not prohibit lebeled statements. :)
Fixed.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/ui/GrayscaleViewModel.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt
alexdtat: Yes, these options are the same. Fixed.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt
alexdtat: > I'm a bit frustrated, probably you can explain this to me. So, we need to do two transformations: rotation and bitmap conversion. Why one of them happens inside `PhotoAnalyzer` and the other one in `ViewModel`?

Rotation when analyzing a photo is needed in order to be able to invoke the functions of the native library from `analyze` with correct images. The `Bitmap` is used only for the GUI, so I decided that it is better to delegate the logic for obtaining it to the `ViewModel`. But, if it's preferable, I can do the `Bitmap` conversion here.

TimPushkin: Ok, this makes sense to me, leave it in the `ViewModel`

app/src/main/java/com/github/featuredetectandroid/ui/theme/FeatureDetectAppTheme.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
alexdtat: Fixed. Also added switching between windos support.

alexdtat:
I hardcoded the camera resolution as 640x360 and removed suppressions for detekt from this PR. In addition, I added support for rotating images 180 and 270 degrees using 90 degree rotations. In the future, it is preferable to replace this with a more efficient solution.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -1,27 +1,136 @@
...
+    }
+
+    private fun cameraPermissionGranted() = ContextCompat.checkSelfPermission(
+        this,
+        Manifest.permission.CAMERA
+    ) == PackageManager.PERMISSION_GRANTED
+
+    private fun shouldRationalize() =
+        (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.M) &&
+            shouldShowRequestPermissionRationale(Manifest.permission.CAMERA)
```
TimPushkin, line 108: If I press `Crtl+Alt+L` Android Studio adds additional indents here

app/src/main/java/com/github/featuredetectandroid/utils/LuminanceArrayToBitmap.kt
```diff
@@ -0,0 +1,36 @@
...
+
+fun luminanceArrayToBitmap(luminanceByteArray: ByteArray, width: Int, height: Int): Bitmap {
+    Log.i(TAG, "Converting ByteArray image of size $width x $height to a Bitmap.")
+
+    val buffer = IntBuffer.allocate(luminanceByteArray.size).apply {
+        for (byte in luminanceByteArray) {
+            val intByte = Y_SHIFT - byte.toInt()
+            put(
+                MAX_ALPHA or (intByte shl RED_SHIFT)
+                    or (intByte shl GREEN_SHIFT) or (intByte shl BLUE_SHIFT)
```
TimPushkin: [According to Android code style](https://developer.android.com/kotlin/style-guide#where_to_break), line break should come after `or`. Also, if I press `Crtl+Alt+L` Android Studio adds additional indents here.

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt
```diff
@@ -0,0 +1,48 @@
...
+    override fun analyze(image: ImageProxy) {
+        val rotationDegrees = image.imageInfo.rotationDegrees
+        Log.i(TAG, "The image is rotated on $rotationDegrees degrees.")
+
+        var width = image.width
+        var height = image.height
+        var oriented = image.planes[0].buffer.toByteArray()
+
+        /** Rotation of incorrectly oriented images is implemented here.
+         * TODO: Optimize rotations for 180 and 270 degrees. */
```
TimPushkin: Why is this a [KDoc](https://kotlinlang.org/docs/kotlin-doc.html#kdoc-syntax) string? I would just place two comments with `//`.

TimPushkin (changes requested):
A few code style changes left and we are done here

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt

app/src/main/java/com/github/featuredetectandroid/utils/LuminanceArrayToBitmap.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
alexdtat: Hm. Ktlint (with `adroid` flag) gives me an error for any other indents.

TimPushkin (approved):

PR #5, opened 2022-07-19
Class architecture,implement SIFT methods, create Test case, include directory i think

zerofivefiveseven:


TimPushkin (changes requested):
Haven't seen the code itself yet, but here are the issues with this PR:
1. It should go into `cpp/main` not `main`.
2. PR name has a typo (no space between "," and "implement"). And why does it end with "I think"? Is it because you aren't sure what you've done?
3. Seems like what you wrote into PR's name should actually be in its description which is currently empty. PR's name should contain the main change you want to merge (maybe its "SIFT implementation" in your case?) and in the description you can place a nicely formatted list of what was done in more details -- just what you wrote in the name of this PR.
4. Commit names are a mess. Firstly, there are three "Add SIFT realization" (by the way, it's usually "implementation" not "realization") with desciption appended one after another -- either rename them to something distinctive or squash them into a single commit. Now it'll require a rather difficult force-push again, but for the future: while the changes are local (not yet pushed) you can just do `git commit --amend` and the current changes will be added to the last commit. Secondly, the last commit starts with "ADD added" for some reason.

PR #6, opened 2022-07-20
Sift implementation

zerofivefiveseven:
Add Class architecture
Implement SIFT methods
Create test case
Add include directory


include/FeatureLib.h
```diff
@@ -0,0 +1,54 @@
...
+#define FEATUREDETECT_FEATURELIB_H_
+
+#include <vector>
+#include <iostream>
+#include <cstdint>
+#include <opencv2/opencv.hpp>
+#include <opencv2/core/core.hpp>
+#include <opencv2/features2d.hpp>
+#include <opencv2/imgproc.hpp>
+namespace libstructs {
```
TimPushkin: Let's add an empty line between the includes and the namespace

include/FeatureLib.h
```diff
@@ -0,0 +1,54 @@
...
+struct KeyPoint {
+  float x;
+  float y;
+  float size;
+  float angle;
+  float response;
+};
+}
```
TimPushkin: Why do you need a separate namespace for this structure? Create a single namespace for the whole project and include the main class in it

include/FeatureLib.h
```diff
@@ -0,0 +1,54 @@
...
+
+class FeatureLib {
+ public:
+  FeatureLib(const std::vector<std::uint8_t> &inputVector, int Width, int Height);
```
TimPushkin: This is a pretty strange variable naming. Rename the variables according to the [code style](https://google.github.io/styleguide/cppguide.html#Variable_Names).

include/FeatureLib.h
```diff
@@ -0,0 +1,54 @@
...
+
+  std::vector<libstructs::KeyPoint> getSiftKeyPoints();
```
TimPushkin: For me "get" refers to something that is either already computed or take little time to compute -- neither of these is our case. Maybe replace "get" with something like "calc" (shortened calculate)?

include/FeatureLib.h
```diff
@@ -0,0 +1,54 @@
...
+
+  std::vector<uint8_t> getSiftDescriptors();
```
TimPushkin: I think, it would be more efficient to calculate key points and descriptors jointly (in a single method) as it would reduce the number of calls through JNI.

Actually, you should ask our main supervisor Anastasiia about it -- she might want to mix key points and descriptors from different algorithms.

include/FeatureLib.h
```diff
@@ -0,0 +1,54 @@
...
+  std::vector<uint8_t> getOrbDescriptors();
+
+  std::vector<libstructs::KeyPoint> getSurfKeyPoints();
+
+  std::vector<uint8_t> getSurfDescriptors();
+
+ private:
+  cv::Mat VectorToMat(const std::vector<std::uint8_t> &inputVector, int Height, int Width);
+  std::vector<uint8_t> MatToVector(const cv::Mat &mat);
+  std::vector<libstructs::KeyPoint> ConvertToStructure(const std::vector<cv::KeyPoint> &CvKeyPointsVec);
```
TimPushkin: Why the public methods have empty lines between them and the private methods don't?

include/FeatureLib.h
```diff
@@ -0,0 +1,54 @@
...
+
+  std::vector<cv::KeyPoint> SiftKeyPoints;
```
TimPushkin: Usually fields are placed before methods

include/FeatureLib.h
```diff
@@ -0,0 +1,54 @@
...
+  std::vector<cv::KeyPoint> OrbKeyPoints;
+  std::vector<cv::KeyPoint> SurfKeyPoints;
+  cv::Mat SiftDescriptors;
+  cv::Mat OrbDescriptors;
+  cv::Mat SurfDescriptors;
+  cv::Mat Image;
+  cv::Ptr<cv::Feature2D> SiftInstance = cv::SIFT::create();
+  cv::Ptr<cv::Feature2D> OrbInstance = cv::ORB::create();
+
```
TimPushkin: Redundant empty line

include/FeatureLib.h
TimPushkin: Rename all the variables in the whole project (including the example) according to the code style, please. Note that there are different rules for different types of variables.

src/featurelib.cpp
```diff
@@ -1,4 +1,49 @@
-#include "featurelib.h"
+#include "FeatureLib.h"
 
-featurelib::featurelib() {
+FeatureLib::FeatureLib(const std::vector<std::uint8_t> &inputVector, int Width, int Height) {
```
TimPushkin: So, one has to reinstantiate the class for each picture? This isn't suitable for real-time calculations.

src/featurelib.cpp
```diff
@@ -1,4 +1,49 @@
...
+  Image = VectorToMat(inputVector, Height, Width);
+  cv::cvtColor( Image, Image, cv::COLOR_BGR2GRAY );
```
TimPushkin: Please press 'Ctrl+Alt+L' here so that this line gets properly formatted

src/featurelib.cpp
```diff
@@ -1,4 +1,49 @@
...
 }
+
+cv::Mat FeatureLib::VectorToMat(const std::vector<std::uint8_t> &inputVector, int Height, int Width) {
+  return cv::Mat(Height, Width, CV_8UC1, (std::uint8_t *) inputVector.data());
```
TimPushkin: Clang-tidy in CLion suggests replacing this with a braced initializer list

TimPushkin: I don't like the C-style cast here. This is something to be avoided in C++.

src/featurelib.cpp
```diff
@@ -1,4 +1,49 @@
...
+}
+
+std::vector<uint8_t> FeatureLib::MatToVector(const cv::Mat &mat) {
```
TimPushkin: Check the mat to vector conversion I did in my project [here](https://github.com/TimPushkin/SgbmDepthAndroid/blob/c2b5e96da5d2fbf49c605e4233c5c8f0318ef606/sgbmdepth/src/main/cpp/src/Conversions.h#L9) and see if its suitable for you. I'm not sure which one is better, so compare them first.

src/featurelib.cpp
```diff
@@ -1,4 +1,49 @@
...
+
+std::vector<libstructs::KeyPoint> FeatureLib::ConvertToStructure(const std::vector<cv::KeyPoint> &CvKeyPointsVec) {
+  std::vector<libstructs::KeyPoint> output;
+  for (auto &kp : CvKeyPointsVec) {
+    libstructs::KeyPoint temp;
+    temp.x = kp.pt.x;
+    temp.y = kp.pt.y;
+    temp.angle = kp.angle;
+    temp.response = kp.response;
+    temp.size = kp.size;
```
TimPushkin: Can't you just initialize the structure using braces? Clang-tidy shows a warning about it being uninitialized anyway.

src/featurelib.cpp
```diff
@@ -1,4 +1,49 @@
...
+  } else {
+    for (int i = 0; i < mat.rows; ++i) {
+      result.insert(result.end(), (uint8_t *) mat.ptr<uchar>(i), (uint8_t *) mat.ptr<uchar>(i) + mat.cols);
+    }
+  }
+  return result;
+}
+
+std::vector<libstructs::KeyPoint> FeatureLib::ConvertToStructure(const std::vector<cv::KeyPoint> &CvKeyPointsVec) {
+  std::vector<libstructs::KeyPoint> output;
```
TimPushkin: I think, it would be more efficient to call [`vector.reserve()`](https://en.cppreference.com/w/cpp/container/vector/reserve) before doing the push backs as you know the size.

examples/test.cpp
```diff
@@ -1,6 +1,33 @@
-#include "../include/featurelib.h"
-
+#include "../include/FeatureLib.h"
+#include <opencv2/core.hpp>
+#include <opencv2/imgcodecs.hpp>
+#include <opencv2/highgui.hpp>
```
TimPushkin: This include is unused

examples/test.cpp
```diff
@@ -1,6 +1,33 @@
...
+std::vector<uint8_t> matToVector(const cv::Mat &mat) {
```
TimPushkin, line 5: Add an empty newline between the includes and the function

examples/test.cpp
```diff
@@ -1,6 +1,33 @@
...
+  std::vector<uint8_t> result;
+  cv::Mat cont = mat.isContinuous() ? mat : mat.clone();  // make sure the Mat is continuous
+  cont.reshape(1, 1).copyTo(result);  // copy from a 1D Mat
+  return result;
+}
 int main() {
```
TimPushkin: Add an empty newline between the functions

examples/test.cpp
```diff
@@ -1,6 +1,33 @@
...
-  featurelib instance;
+  std::string image_path = cv::samples::findFile("church.jpg");
```
TimPushkin: It would be better to accept an input from command line and pick a file chosen by a user

examples/test.cpp
```diff
@@ -1,6 +1,33 @@
...
+  FeatureLib featurelib(testVector, width, height);
+
+  std::vector<libstructs::KeyPoint> testVectorOfKeyPoints = featurelib.getSiftKeyPoints();
+  std::vector<uint8_t> testVectorOfDescriptors = featurelib.getSiftDescriptors();
+  std::cout << "Printing points" << std::endl;
+  for (auto &tempKp : testVectorOfKeyPoints) {
+    std::cout << tempKp.x << "\t" << tempKp.y << std::endl;
+  }
+  std::cout << "Printing descriptors" << std::endl;
+
```
TimPushkin, line 36: This is a strange newline. Wouldn't it be better to separate the calculations, printing, and returning?

TimPushkin (changes requested):
General notes:
1. Can you add a `.gitignore` for `.idea` and `build` dirs?
2. Why the source file is called `featurelib.cpp` while the corresponding header is called `FeatureLib.h`? Read what [the code style has to say about it](https://github.com/TimPushkin/FeatureDetectAndroid.git) and name both files accordingly.

Architecture:
1. Main class has a lot of conversions which can be made static. You can move them to a separate file, maybe even a header if some of them will become templates after applying my comments.
2. We need to calculate key points and descriptors in real-time and our main use case implies sending a new picture each time, so you need to take a picture as an argument in your methods, not in the class constructor.
3. As you won't save the picture in a class'es field, you can create a separate class for each algorithm and let them have the same interface.
4. Don't use class fields for key points and descriptors -- just create local variables. Even if you won't unite the methods for calculating them, it's a user's duty to save key points and pass them to you as arguments to calculate descriptors for them. But I think, these methods will be united anyway.

zerofivefiveseven:
General notes:
1.Done
2.Can you send me link again, link in summary isn't working, but i renamed them anyway

zerofivefiveseven:
Architecture:
>>Main class has a lot of conversions which can be made static. You can move them to a separate file, maybe even a header if some of them will become templates after applying my comments.
From [stackoverflow:](https://stackoverflow.com/questions/31305717/member-function-with-static-linkage)
_"In the context of member functions, static means "this member function does not have a receiver object. It's basically a normal function that's nested inside of the scope of the class."_

If i declare them into my class.h, then when i initialize them in conversions.cpp i should type static or not?

zerofivefiveseven:
Architecture:
>As you won't save the picture in a class'es field, you can create a separate class for each algorithm and let them have the same interface.
I don't understand how I should do it. [Nested classes](https://www.tutorialspoint.com/nested-classes-in-cplusplus#:~:text=A%20nested%20class%20is%20a,members%20of%20a%20nested%20class.)? Or other architecture practice?

include/FeatureLib.h
zerofivefiveseven: done

include/FeatureLib.h
zerofivefiveseven: done

include/FeatureLib.h
zerofivefiveseven: done

include/FeatureLib.h
zerofivefiveseven: done, now there is a separator line between private methods declaration

src/featurelib.cpp
zerofivefiveseven: done

src/featurelib.cpp
zerofivefiveseven: done

src/featurelib.cpp
zerofivefiveseven: In my case, function described in your project don't work, some errors while building, i'll send you log today

src/featurelib.cpp
zerofivefiveseven: done

src/featurelib.cpp
zerofivefiveseven: done
but don't know how to get number of structure fields by method, or something like that

examples/test.cpp
zerofivefiveseven: done

examples/test.cpp
zerofivefiveseven: done

examples/test.cpp
zerofivefiveseven: done

examples/test.cpp
zerofivefiveseven: done

TimPushkin:
> Can you send me link again, link in summary isn't working, but i renamed them anyway

Yeah, sorry, here it is: https://google.github.io/styleguide/cppguide.html#Variable_Names

> If i declare them into my class.h, then when i initialize them in conversions.cpp i should type static or not?

I would consult CLion for this :)
But why declaring them in `class.h` if they can be declared (and, actually, defined if needed) in a separate header?

> I don't understand how I should do it. Nested classes? Or other architecture practice?

No, I mean just create three independent classes which will inherit from some common interface class with virtual methods.

src/featurelib.cpp
TimPushkin: Some problems may arise if you place it in a source instead of a header file because my function is a template

src/featurelib.cpp
TimPushkin: Sorry, I don't understand the question

include/FeatureLib.h
zerofivefiveseven: done

include/FeatureLib.h
zerofivefiveseven: asked

examples/test.cpp
zerofivefiveseven: done, but now my test fails:
cv::cvtColor(Image, Image, cv::COLOR_BGR2GRAY);  with that line test isn't working
> Invalid number of channels in input image:
>     'VScn::contains(scn)'
> where
>     'scn' is 1

src/featurelib.cpp
zerofivefiveseven: Yep, my bad
yep, i should handle with templates and abstract methods and static methods.
Tonight, i think

examples/test.cpp
TimPushkin: You can debug this by checking Mat type of `Image`. You'll probably need this: https://gist.github.com/yangcha/38f2fa630e223a8546f9b48ebbb3e61a

zerofivefiveseven: Yep, i fix it

include/FeatureLib.h
zerofivefiveseven: done

src/featurelib.cpp
zerofivefiveseven: can't handle with that

PR #8, opened 2022-07-21
Refactor classes

zerofivefiveseven:
wrote some methods to conversions.h
new structure
new conversions method
new base class
refactor calc methods


include/feature_lib.h
```diff
@@ -0,0 +1,49 @@
+#ifndef FEATUREDETECT_FEATURELIB_H_
+#define FEATUREDETECT_FEATURELIB_H_
+
+#include <vector>
+#include <iostream>
+#include <cstdint>
+#include <opencv2/opencv.hpp>
+#include <opencv2/core/core.hpp>
+#include <opencv2/features2d.hpp>
+#include <opencv2/imgproc.hpp>
```
TimPushkin: Put as much includes in the source file and leave here only those that are used in this header directly. You can find more information why [here](https://stackoverflow.com/questions/3002110/include-in-h-or-c-cpp).

include/feature_lib.h
```diff
@@ -0,0 +1,49 @@
...
+
+namespace libstructs {
```
TimPushkin: Let's name it `featurelib` as it contains the whole library

include/feature_lib.h
```diff
@@ -0,0 +1,49 @@
...
+
+struct KeyPoint {
+  float x;
+  float y;
+  float size;
+  float angle;
+  float response;
```
TimPushkin: Let's name it `strength` as nobody known what does "response" mean. Different implementations name it differently anyway.

include/feature_lib.h
```diff
@@ -0,0 +1,49 @@
...
+};
+
+struct CalcOutputStruct {
```
TimPushkin: No point in naming a `struct` a Struct. Let's name it `DetectionResult`.

include/feature_lib.h
```diff
@@ -0,0 +1,49 @@
...
+  std::vector<KeyPoint> keypoints;
+  std::vector<uint8_t> descriptors;
+};
+
+class FeatureDetector {
+ public:
+  // pure virtual function providing interface framework.
```
TimPushkin: Seems too low level, let's remove this doc string.

include/feature_lib.h
```diff
@@ -0,0 +1,49 @@
...
+  virtual CalcOutputStruct calc(const std::vector<std::uint8_t> &input_vector) = 0;
```
TimPushkin: Now I think, maybe "detect" would be better as the class is a detector? If you have some other thoughts, feel free to share as I'm not sure here.

include/feature_lib.h
```diff
@@ -0,0 +1,49 @@
...
+
```
TimPushkin, line 25: Remove this blank line

include/feature_lib.h
```diff
@@ -0,0 +1,49 @@
...
+};
+
+class SiftDetector : public FeatureDetector {
+ public:
+  SiftDetector(int width, int height);
+
+  CalcOutputStruct calc(const std::vector<std::uint8_t> &input_vector);
```
TimPushkin: Clang-tidy suggests annotating it with `override`. Don't you have clang-tidy enabled in CLion?

include/feature_lib.h
```diff
@@ -0,0 +1,49 @@
...
+
+ private:
+  int width;
+  int height;
+
+  cv::Ptr<cv::Feature2D> sift_instance_ = cv::SIFT::create();
```
TimPushkin: I think, calling it just `sift_` would be shorter and still understandable

include/feature_lib.h
```diff
@@ -0,0 +1,49 @@
...
+};
+
+}  //namespace libstructs
```
TimPushkin: It would be better to add a space between `//` and `namespace` if the linter doesn't mind

src/feature_lib.cpp
```diff
@@ -0,0 +1,23 @@
+#include "feature_lib.h"
+#include "conversions.h"
+namespace libstructs {
```
TimPushkin: Add a blank line here

src/feature_lib.cpp
```diff
@@ -0,0 +1,23 @@
...
+  std::vector<cv::KeyPoint> SiftKeyPoints;
+  cv::Mat SiftDescriptors;
+
+  sift_instance_->detectAndCompute(Image, cv::Mat(), SiftKeyPoints, SiftDescriptors);
+  return ConvertToOutputStructure(ConvertToStructure(SiftKeyPoints), MatToVector(SiftDescriptors));
+}
+
+}  // namespace libstructs
+
+
```
TimPushkin, line 43: Three blank lines is too much, a single one is enough :)

src/feature_lib.cpp
```diff
@@ -0,0 +1,23 @@
...
+namespace libstructs {
+
+SiftDetector::SiftDetector(int width, int height) {
+  this->width = width;
+  this->height = height;
+}
+
+libstructs::CalcOutputStruct SiftDetector::calc(const std::vector<std::uint8_t> &input_vector) {
+  cv::Mat Image = VectorToMat(input_vector, height, width);
+  cv::cvtColor(Image, Image, cv::COLOR_BGR2GRAY);  //with that line test isn't working
```
TimPushkin: I don't think this comment is still relevant

TimPushkin: We should probably do `RGB2GRAY` instead as it is much more popular nowadays

TimPushkin: You may also need to tweak your test (by adding a `BGR2RGB` conversion) after this because OpenCV would read an image in BGR by default, I believe

src/feature_lib.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+  std::vector<cv::KeyPoint> SiftKeyPoints;
+  cv::Mat SiftDescriptors;
```
TimPushkin: These names don't match our code style

src/feature_lib.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+  sift_instance_->detectAndCompute(Image, cv::Mat(), SiftKeyPoints, SiftDescriptors);
```
TimPushkin: Are you sure that passing an uninitialized `cv::Mat()` as a mask is ok? Shouldn't it be initialized to all 255 or something? It would be great if you could provide a link to the docs or an example.

src/feature_lib.cpp
```diff
@@ -0,0 +1,23 @@
...
+#include "conversions.h"
+namespace libstructs {
+
+SiftDetector::SiftDetector(int width, int height) {
+  this->width = width;
+  this->height = height;
+}
+
+libstructs::CalcOutputStruct SiftDetector::calc(const std::vector<std::uint8_t> &input_vector) {
+  cv::Mat Image = VectorToMat(input_vector, height, width);
```
TimPushkin: This name doesn't match the code style

src/conversions.h
```diff
@@ -0,0 +1,39 @@
+#include <vector>
+#include <iostream>
+#include <cstdint>
+#include <opencv2/opencv.hpp>
+#include <opencv2/core/core.hpp>
+#include "feature_lib.h"
```
TimPushkin: I believe, includes should go inside the include guards (i.e. after `ifndef` and `define`)

src/conversions.h
```diff
@@ -0,0 +1,39 @@
...
+
+#ifndef FEATUREDETECT_SRC_CONVERSIONS_H_
+#define FEATUREDETECT_SRC_CONVERSIONS_H_
+cv::Mat VectorToMat(const std::vector<std::uint8_t> &inputVector, int Height, int Width) {
```
TimPushkin: Add a blank line

TimPushkin: This function name and the names of some arguments don't match the code style

TimPushkin: Actually, all function names in this file don't, so fix them all, please

src/conversions.h
```diff
@@ -0,0 +1,39 @@
...
+
+std::vector<uint8_t> MatToVector(const cv::Mat &mat) {
+  std::vector<uint8_t> result;
+  cv::Mat cont = mat.isContinuous() ? mat : mat.clone();  // make sure the Mat is continuous
+  cont.reshape(1, 1).copyTo(result);  // copy from a 1D Mat
+
+  return result;
+}
+
+std::vector<libstructs::KeyPoint> ConvertToStructure(const std::vector<cv::KeyPoint> &cv_key_points_vec) {
```
TimPushkin: I think, naming the argument as `cv_key_points` would be enough

src/conversions.h
```diff
@@ -0,0 +1,39 @@
...
+  std::vector<libstructs::KeyPoint> output;
+  for (auto &kp : cv_key_points_vec) {
+    libstructs::KeyPoint temp = {kp.pt.x, kp.pt.y, kp.angle, kp.response, kp.size};
+    output.reserve(5);
```
TimPushkin: You can call `output.reserve(cv_key_points_vec.size())` once before the cycle -- this would be more efficient

TimPushkin: And do you really need to allocate five elements to push a single one? I strongly believe, it takes the type and its size into account.

src/conversions.h
```diff
@@ -0,0 +1,39 @@
...
+    output.push_back(temp);
+  }
+  return output;
+}
+
+libstructs::CalcOutputStruct ConvertToOutputStructure(const std::vector<libstructs::KeyPoint> cv_key_points_vec,
```
TimPushkin: These are our converted key points, not the ones from OpenCV. So, you can rename the parameter to `key_points`.

src/conversions.h
```diff
@@ -0,0 +1,39 @@
...
+                                                      const std::vector<uint8_t> descriptors) {
```
TimPushkin: Clang-tidy suggests making both of these parameters references

src/conversions.h
TimPushkin: Do we need this function at all? It can be replaced with `CalcOutputStruct {key_points, descriptors}`.

src/conversions.h
```diff
@@ -0,0 +1,39 @@
...
+  libstructs::CalcOutputStruct output;
+  output.keypoints = cv_key_points_vec;
+  output.descriptors = descriptors;
+  return output;
+}
+#endif //FEATUREDETECT_SRC_CONVERSIONS_H_
```
TimPushkin: Add a blank line here

TimPushkin: Add a space between `//` and `FEATUREDETECT_SRC_CONVERSIONS_H_`

examples/test.cpp
```diff
@@ -1,6 +1,43 @@
...
+#include <opencv2/imgcodecs.hpp>
+
+std::vector<uint8_t> matToVector(const cv::Mat &mat) {
+  std::vector<uint8_t> result;
+  cv::Mat cont = mat.isContinuous() ? mat : mat.clone();  // make sure the Mat is continuous
+  cont.reshape(1, 1).copyTo(result);  // copy from a 1D Mat
+  return result;
+}
+
+int main(int argc, char *argv[]) {
```
TimPushkin: Add some docs describing what a user has to pass as the CLI arguments

examples/test.cpp
```diff
@@ -1,6 +1,43 @@
...
+
```
TimPushkin: Redundant blank line

examples/test.cpp
```diff
@@ -1,6 +1,43 @@
...
+  if (argc != 2) {
+    std::cerr << "Illegal CLI arguments num. Expected only filepath (including the command name) but was "
```
TimPushkin: `std::cerr << "Illegal CLI arguments num. Expected only a file path but was " << argc << " arguments (including the command name)"` would be better

examples/test.cpp
```diff
@@ -1,6 +1,43 @@
...
+              << argc;
```
TimPushkin: Currently it can fit on a single line

examples/test.cpp
```diff
@@ -1,6 +1,43 @@
...
+    return 1;
+  }
+  std::string file_path = argv[1];
+  std::string image_path = cv::samples::findFile(file_path);
```
TimPushkin: What do you need this line for?

examples/test.cpp
```diff
@@ -1,6 +1,43 @@
...
+  cv::Mat img = imread(file_path, cv::IMREAD_COLOR);
+
+  int width = img.cols;
+  int height = img.rows;
+  libstructs::SiftDetector sift_example(width, height);
```
TimPushkin: Let's name it "detector"

examples/test.cpp
```diff
@@ -1,6 +1,43 @@
...
+
+  std::vector<uint8_t> test_vector = matToVector(img);
```
TimPushkin: Let's name it "image_data"

examples/test.cpp
```diff
@@ -1,6 +1,43 @@
...
+
+  libstructs::SiftDetector instance(width, height);
```
TimPushkin: Why do you create a second instance of the detector here?

examples/test.cpp
TimPushkin: Actually, you create an instance of this class again later

examples/test.cpp
```diff
@@ -1,6 +1,43 @@
...
+  libstructs::CalcOutputStruct test_output = instance.calc(test_vector);
```
TimPushkin: Let's call it "output"

examples/test.cpp
```diff
@@ -1,6 +1,43 @@
...
+
+  std::cout << "Printing points" << std::endl;
+  for (auto &temp_kp : test_output.keypoints) {
```
TimPushkin: "kp" instead of "temp_kp" would be shorter and still understandable

examples/test.cpp
```diff
@@ -1,6 +1,43 @@
...
+    std::cout << temp_kp.x << "\t" << temp_kp.y << std::endl;
+  }
+
+  std::cout << "Printing descriptors" << std::endl;
+  for (auto &temp_uint : test_output.descriptors) {
```
TimPushkin: Let's call it "desc_part". And, when you fix the descriptors array, print each descriptor on a separate line with parts saparated with spaces

examples/test.cpp
```diff
@@ -1,6 +1,43 @@
...
+  libstructs::SiftDetector sift_example(width, height);
+
+  std::vector<uint8_t> test_vector = matToVector(img);
+
+  libstructs::SiftDetector instance(width, height);
+  libstructs::CalcOutputStruct test_output = instance.calc(test_vector);
+
+  std::cout << "Printing points" << std::endl;
+  for (auto &temp_kp : test_output.keypoints) {
+    std::cout << temp_kp.x << "\t" << temp_kp.y << std::endl;
```
TimPushkin: Let's print all parts of a key point including size, angle, etc.

TimPushkin (changes requested):
Looks great overall, but descriptors should be represented by a `vector<vector<uint8_t>>` because each key point has an assiciated array of `uint8_t`s.

TimPushkin:
Also, from the architectural aspect every class and struct in `feature_lib.h` should've been placed into a separate file. But as far as I know, Scapix has problems with porting structs from different headers, so we have to place everything together.

include/feature_lib.h
TimPushkin: And let's call the argument `image`

include/feature_lib.h
zerofivefiveseven: done

zerofivefiveseven: read

include/feature_lib.h
zerofivefiveseven: done

include/feature_lib.h
zerofivefiveseven: done

include/feature_lib.h
zerofivefiveseven: done

include/feature_lib.h
zerofivefiveseven: done

include/feature_lib.h
zerofivefiveseven: done

include/feature_lib.h
zerofivefiveseven: done
I just didn't pay attention
before i start work at this pr i notice that

include/feature_lib.h
```diff
@@ -0,0 +1,49 @@
...
+  float response;
+};
+
+struct CalcOutputStruct {
+  std::vector<KeyPoint> keypoints;
+  std::vector<uint8_t> descriptors;
+};
+
+class FeatureDetector {
+ public:
```
TimPushkin, line 23: You should provide setters (and getters) for `width` and `height` so that a user won't have to reinitialize the whole class to change them

include/feature_lib.h
zerofivefiveseven: done

include/feature_lib.h
zerofivefiveseven: done

src/feature_lib.cpp
zerofivefiveseven: done

src/feature_lib.cpp
zerofivefiveseven: done

src/feature_lib.cpp
zerofivefiveseven: done

src/feature_lib.cpp
zerofivefiveseven: done

src/feature_lib.cpp
zerofivefiveseven: https://stackoverflow.com/questions/16365129/opencv-howto-use-mask-parameter-for-feature-point-detection-surf
I think it should be zero matrix

zerofivefiveseven: with our image shape

src/feature_lib.cpp
zerofivefiveseven: done

src/conversions.h
zerofivefiveseven: done

src/conversions.h
zerofivefiveseven: done

src/conversions.h
zerofivefiveseven: done

include/feature_lib.h
zerofivefiveseven: I think that we calculate them, so 'calc' is what we need

src/feature_lib.cpp
zerofivefiveseven: in work
deal with this tonight as a last priority

src/feature_lib.cpp
zerofivefiveseven: https://docs.opencv.org/3.4/d0/d13/classcv_1_1Feature2D.html
detect() method.
when i fix how image sizes kept in our class i ll deal with that

src/conversions.h
zerofivefiveseven: done. i just I just didn't understand, now I understand

src/conversions.h
zerofivefiveseven: done

src/conversions.h
zerofivefiveseven: done

include/feature_lib.h
TimPushkin: But the process is actually called [feature detection](https://en.wikipedia.org/wiki/Feature_(computer_vision)), so we _detect_ them indeed

src/feature_lib.cpp
TimPushkin: Seems like it should be a non-zero (255 probably) matrix sized the same as the image. See [here](https://answers.opencv.org/question/103087/behaviour-of-orb-keypoint-detection-with-masks/).

examples/test.cpp
zerofivefiveseven: remove that and use previous one

zerofivefiveseven: done

examples/test.cpp
zerofivefiveseven: done

examples/test.cpp
zerofivefiveseven: done

examples/test.cpp
zerofivefiveseven: done

src/conversions.h
zerofivefiveseven: >Actually, all function names in this file don't, so fix them all, please
public:
  int CountFooErrors(const std::vector<Foo>& foos) {
that example from google C++ codestyle link
VectorToMat matches, isn't it?

src/conversions.h
zerofivefiveseven: done

zerofivefiveseven: done

examples/test.cpp
zerofivefiveseven: do

examples/test.cpp
zerofivefiveseven: done

examples/test.cpp
zerofivefiveseven: done

examples/test.cpp
zerofivefiveseven: done

examples/test.cpp
zerofivefiveseven: okay, done

examples/test.cpp
zerofivefiveseven: done

examples/test.cpp
zerofivefiveseven: done

src/feature_lib.cpp
zerofivefiveseven: nope, tomorrow

src/conversions.h
TimPushkin: Yep, you are right here about the function names

TimPushkin: Actually, let's make an exception here and name functions in `camelCase` with leading lowercase letter, because we intend to generate an interface for Java where all class members will have the same names as they do in C++.

include/feature_lib.h
zerofivefiveseven: done

src/conversions.h
zerofivefiveseven: okay, should i rename whole project functions? or only in conversions file?

include/feature_lib.h
zerofivefiveseven: done

examples/test.cpp
zerofivefiveseven: deleted. that line uses for "smart search" but as the result it wasn't smart at all

zerofivefiveseven: done

src/conversions.h
zerofivefiveseven: done

src/feature_lib.cpp
zerofivefiveseven: done

src/feature_lib.cpp
zerofivefiveseven: i'll test it out
now RGB2GRAY flag

src/conversions.h
TimPushkin: In fact, only the public ones will be ported, but let's rename all of them for the uniformity

examples/test.cpp
TimPushkin: Nope, it's still called `test_output`

include/feature_lib.h
```diff
@@ -0,0 +1,59 @@
...
+};
+
+struct DetectionResult {
+  std::vector<KeyPoint> keypoints;
+  std::vector<std::vector<uint8_t>> descriptors;
+};
+
+class FeatureDetector {
+ public:
+  virtual DetectionResult detect(const std::vector<std::uint8_t> &input_vector) = 0;
```
TimPushkin: I think, this function can be made `const` too

include/feature_lib.h
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+ protected:
+  int width;
+  int height;
+};
+
+class SiftDetector : public FeatureDetector {
+ public:
+  SiftDetector();
```
TimPushkin: What do you need this constructor for? A default one is generated automatically. Though, I would suggest adding another constructor (and keep this one too) with two parameters: `width` and `height` -- set the according fields to the received values so that one won't have to call the corresponding setters.

TimPushkin: After you add such constructor, you can simplify your test

include/feature_lib.h
```diff
@@ -0,0 +1,59 @@
...
+
+  DetectionResult detect(const std::vector<std::uint8_t> &input_vector) override;
+
+ private:
+  cv::Ptr<cv::Feature2D> sift_ = cv::SIFT::create();
+};
+
+}  // namespace featurelib
```
TimPushkin: Remove a redundant space before `//` -- one is enough

src/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-set(SOURCE_FILES featurelib.cpp)
+set(SOURCE_FILES feature_lib.cpp conversions.h)
```
TimPushkin: You don't need to put headers into sources

src/conversions.h
```diff
@@ -0,0 +1,44 @@
+#ifndef FEATUREDETECT_SRC_CONVERSIONS_H_
+#define FEATUREDETECT_SRC_CONVERSIONS_H_
+
+#include <vector>
+#include <iostream>
+#include <cstdint>
+#include <opencv2/opencv.hpp>
+#include <opencv2/core/core.hpp>
+#include "feature_lib.h"
```
TimPushkin, line 5: Some includes here are unused

src/conversions.h
```diff
@@ -0,0 +1,44 @@
...
+
+cv::Mat vectorToMat(const std::vector<std::uint8_t> &input_vector, int height, int width) {
+  return {height, width, CV_8UC3, (std::uint8_t *) input_vector.data()};
+}
+
+std::vector<std::vector<uint8_t> > matTo2DVec(const cv::Mat &mat_in) {
```
TimPushkin: Why the function above uses "vector" and this one uses "vec"? Choose one of them for consictency.

TimPushkin: Let's call the argument "mat" as it's shorter

src/conversions.h
```diff
@@ -0,0 +1,44 @@
...
+  std::vector<std::vector<uint8_t> > vec_out(mat_in.rows);
+  for (int i = 0; i < mat_in.rows; ++i) {
+    vec_out[i].resize(mat_in.cols);
+    for (int j = 0; j < mat_in.cols; ++j) {
+      vec_out[i][j] = mat_in.at<uint8_t>(i, j);
+    }
+  }
+  return vec_out;
```
TimPushkin: I think, you can do this conversion faster like [here](https://github.com/TimPushkin/SgbmDepthAndroid/blob/c2b5e96da5d2fbf49c605e4233c5c8f0318ef606/sgbmdepth/src/main/cpp/src/Conversions.h#L17) (I don't know why it is a `template` though as you need to define a concrete type -- change `float` to `uint8_t` in your case).

src/conversions.h
```diff
@@ -0,0 +1,44 @@
...
+}
+
+std::vector<uint8_t> matTo1DVector(const cv::Mat &mat) {
```
TimPushkin: Seems like this is not needed anymore

src/conversions.h
```diff
@@ -0,0 +1,44 @@
...
+  std::vector<featurelib::KeyPoint> output;
+  output.reserve(key_points.size());
+  for (auto &kp : key_points) {
+    featurelib::KeyPoint temp = {kp.pt.x, kp.pt.y, kp.angle, kp.response, kp.size};
+    output.push_back(temp);
+  }
+  return output;
+}
+
+#endif //  FEATUREDETECT_SRC_CONVERSIONS_H_
```
TimPushkin: Remove a redundant space after `//` -- one is enough

src/feature_lib.cpp
```diff
@@ -0,0 +1,21 @@
+#include "feature_lib.h"
+#include "conversions.h"
+
+namespace featurelib {
+
+SiftDetector::SiftDetector() {}
```
TimPushkin: Clang-tidy suggests using `default` here

src/feature_lib.cpp
```diff
@@ -0,0 +1,21 @@
...
+
+featurelib::DetectionResult SiftDetector::detect(const std::vector<std::uint8_t> &input_vector) {
```
TimPushkin: Let's call the parameter "input" -- no need to repeat its type in its name

src/feature_lib.cpp
```diff
@@ -0,0 +1,21 @@
...
+  cv::Mat image = vectorToMat(input_vector, height, width);
+  cv::cvtColor(image, image, cv::COLOR_RGB2GRAY);
+
+  std::vector<cv::KeyPoint> sift_key_points;
+  cv::Mat sift_descriptors;
+  cv::Mat mask(255 * cv::Mat::ones(height, width,CV_8U));
```
TimPushkin: You can put it in class field so it won't be recreated for each image. And don't forget to update its sizes in width and height setters.

src/feature_lib.cpp
```diff
@@ -0,0 +1,21 @@
...
+  sift_->detectAndCompute(image, mask, sift_key_points, sift_descriptors);
+  std::vector<KeyPoint> key_points_output = convertToStructure(sift_key_points);
```
TimPushkin: Let's call it "key_points"

src/feature_lib.cpp
```diff
@@ -0,0 +1,21 @@
...
+  std::vector<std::vector<uint8_t>> descriptors_output = matTo2DVec(sift_descriptors);
```
TimPushkin: Let's call it "descriptors"

src/feature_lib.cpp
```diff
@@ -0,0 +1,21 @@
...
+  return {key_points_output, descriptors_output};
```
TimPushkin: Add some blank lines separating code blocks logically -- it's not easy to read seven different lines without any separation.

src/feature_lib.cpp
```diff
@@ -0,0 +1,21 @@
...
+}
+
+}  // namespace featurelib
```
TimPushkin: Remove the redundant space before `//` -- one is enough

examples/test.cpp
```diff
@@ -1,6 +1,43 @@
-#include "../include/featurelib.h"
+#include "../include/feature_lib.h"
+#include <iostream>
+#include <opencv2/core.hpp>
+#include <opencv2/imgcodecs.hpp>
```
TimPushkin: Rearrange includes as the code style [suggests](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes)

examples/test.cpp
```diff
@@ -1,6 +1,43 @@
...
 
-int main() {
-  featurelib instance;
+std::vector<uint8_t> matToVector(const cv::Mat &mat) {
+  std::vector<uint8_t> result;
+  cv::Mat cont = mat.isContinuous() ? mat : mat.clone();  // make sure the Mat is continuous
+  cont.reshape(1, 1).copyTo(result);  // copy from a 1D Mat
+  return result;
+}
+//Please inter only filepath to command line arguments
```
TimPushkin: Add a blank line between `matToVector` and this doc

TimPushkin: 1. Add a space between `//` and the comment itself.
2. Fix typos: it's "enter" not "inter".
3. You don't write docs like this: no need to write "please" -- you describe what the function does, what args it accepts, and what it returns. And you make it as short as possible. You can see some examples [in the code style](https://google.github.io/styleguide/cppguide.html#Function_Comments).

examples/test.cpp
```diff
@@ -1,6 +1,43 @@
...
+int main(int argc, char *argv[]) {
+  if (argc != 2) {
+    std::cerr << "Illegal CLI arguments num. Expected only a file path but was " << argc << " arguments (including the command name)"<<std::endl;
```
TimPushkin: Now this line exceeds the length limit. Press `Ctrl+Alt+L` and CLion will fix this for you.

examples/test.cpp
```diff
@@ -1,6 +1,43 @@
...
+  cv::Mat img = imread(file_path, cv::IMREAD_COLOR);
+
+  int width = img.cols;
+  int height = img.rows;
+
+  std::vector<uint8_t> image_data = matToVector(img);
+
+  featurelib::SiftDetector detector;
+  detector.setWidth(width);
+  detector.setHeight(height);
```
TimPushkin: This setters will be gone after you add a new constructor to the detector class

examples/test.cpp
zerofivefiveseven: done

examples/test.cpp
```diff
@@ -1,6 +1,43 @@
...
+  for (auto &kp : test_output.keypoints) {
+    std::cout << kp.x << "\t" << kp.y << "\t" << kp.size << "\t" << kp.angle << "\t" << kp.strength << std::endl;
+  }
+
+  std::cout << "Printing descriptors" << std::endl;
+  for ( const auto &row : test_output.descriptors )
+  {
+    for ( const auto &s : row ) std::cout << s << ' ';
+    std::cout << std::endl;
+  }
```
TimPushkin: Wrong formatting here, press `Ctrl+Alt+L` to fix this

examples/test.cpp
```diff
@@ -1,6 +1,43 @@
...
+
+  std::cout << "Printing points" << std::endl;
+  for (auto &kp : test_output.keypoints) {
+    std::cout << kp.x << "\t" << kp.y << "\t" << kp.size << "\t" << kp.angle << "\t" << kp.strength << std::endl;
+  }
+
+  std::cout << "Printing descriptors" << std::endl;
+  for ( const auto &row : test_output.descriptors )
+  {
+    for ( const auto &s : row ) std::cout << s << ' ';
```
TimPushkin: What does `s` mean? Pick a name that makes more sense.

TimPushkin (changes requested):
1. Use `auto` where possible if it doesn't make the code less clear. You can read more about when to use `auto` [here](https://google.github.io/styleguide/cppguide.html#Type_deduction).
2. Let's rename `test.cpp` to `detection_example.cpp`.

include/feature_lib.h
zerofivefiveseven: done

include/feature_lib.h
zerofivefiveseven: done

include/feature_lib.h
zerofivefiveseven: done

src/CMakeLists.txt
zerofivefiveseven: done

src/conversions.h
zerofivefiveseven: done

include/feature_lib.h
zerofivefiveseven: renamed to vector_out

src/conversions.h
zerofivefiveseven: renamed to vector_out

zerofivefiveseven: done

src/conversions.h
zerofivefiveseven: done, I didn't delete the template, I decided that it would come later

src/conversions.h
zerofivefiveseven: done

src/conversions.h
zerofivefiveseven: done

src/conversions.h
zerofivefiveseven: okay, it doesn't work, remove template

src/feature_lib.cpp
zerofivefiveseven: done

src/feature_lib.cpp
zerofivefiveseven: done

src/feature_lib.cpp
zerofivefiveseven: done

src/feature_lib.cpp
zerofivefiveseven: done

src/feature_lib.cpp
zerofivefiveseven: done

src/feature_lib.cpp
zerofivefiveseven: done

src/feature_lib.cpp
zerofivefiveseven: done

examples/test.cpp
zerofivefiveseven: done

examples/test.cpp
zerofivefiveseven: done

examples/test.cpp
zerofivefiveseven: done

examples/test.cpp
zerofivefiveseven: done

examples/test.cpp
zerofivefiveseven: done

examples/test.cpp
zerofivefiveseven: done

examples/test.cpp
zerofivefiveseven: done

src/conversions.h
TimPushkin: But why `vector_out`? I was talking about the function name, thought it would be just `matTo2DVector`.

zerofivefiveseven: so i renamed func too

src/feature_lib.cpp
TimPushkin: Nope, still `key_points_output`. You acceentally renamed `sift_key_points` -- leave them as they were back then.

src/feature_lib.cpp
TimPushkin: Nope, still `descriptors_output`. You acceentally renamed `sift_descriptors` -- leave them as they were back then.

examples/detection_example.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+std::vector<uint8_t> matToVector(const cv::Mat &mat) {
+  std::vector<uint8_t> result;
+  cv::Mat cont = mat.isContinuous() ? mat : mat.clone();  // make sure the Mat is continuous
+  cont.reshape(1, 1).copyTo(result);  // copy from a 1D Mat
+  return result;
+}
+
+// Calculates key points and image descriptors
+// Takes a filepath as input
```
TimPushkin: "image file path" would be better

examples/detection_example.cpp
```diff
@@ -0,0 +1,45 @@
...
+// Returns data in the structure declared in associated header
```
TimPushkin: It actually returns nothing, it prints keypoints and descriptors to `stdout`

include/feature_lib.h
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  int getHeight() const {
+    return height;
+  }
+
+ protected:
+  int width;
+  int height;
+  cv::Mat mask = cv::Mat(255 * cv::Mat::ones(height, width, CV_8U));
```
TimPushkin: Bring back the setters. We want to allow our users to change these without recreating an object. And don't forget to recalculate the mask in these setters.

include/feature_lib.h
```diff
@@ -0,0 +1,53 @@
...
+};
+
+class SiftDetector : public FeatureDetector {
+ public:
+  SiftDetector();
+  SiftDetector(int width, int height);
```
TimPushkin, line 46: Add a blank line to separate the constructors

include/feature_lib.h
```diff
@@ -0,0 +1,53 @@
...
+
+  DetectionResult detect(const std::vector<std::uint8_t> &input) const;
```
TimPushkin: Bring back the `override` specifier. Clang-tidy suggests it -- pay attention to its warnings.

src/conversions.h
```diff
@@ -0,0 +1,36 @@
+#ifndef FEATUREDETECT_SRC_CONVERSIONS_H_
+#define FEATUREDETECT_SRC_CONVERSIONS_H_
+
+#include <opencv2/opencv.hpp>
```
TimPushkin: I think, you need `opencv2/core.hpp` here

src/conversions.h
```diff
@@ -0,0 +1,36 @@
...
+std::vector<std::vector<uint8_t>> matTo2DVector(const cv::Mat &mat) {
+  std::vector<std::vector<uint8_t>> result;
+  cv::Mat cont = mat.isContinuous() ? mat : mat.clone();  // make sure the Mat is continuous
+
+  result.reserve(cont.rows);
+
+  auto rowLen = cont.cols * cont.channels();
+  for (auto i = 0; i < cont.rows; i++) {
+    const auto *rowPtr = cont.ptr<uint8_t>(i);
+    result.push_back(std::vector<uint8_t>(rowPtr, rowPtr + rowLen));
```
TimPushkin: Clang-tidy suggests using `emplace_back` here

src/feature_lib.cpp
```diff
@@ -0,0 +1,28 @@
+#include "feature_lib.h"
+#include "conversions.h"
```
TimPushkin, line 3: You need to include `opencv2/imgproc.hpp` -- currently you are getting `cv::COLOR_RGB2GRAY` through a transitive dependency from `conversions.h` which is bad

TimPushkin (changes requested):
You can still add `auto` in a lot of places

examples/detection_example.cpp
zerofivefiveseven: done

examples/detection_example.cpp
zerofivefiveseven: done

include/feature_lib.h
zerofivefiveseven: done

include/feature_lib.h
zerofivefiveseven: done

src/conversions.h
zerofivefiveseven: it still building without, but i ll do that

src/conversions.h
zerofivefiveseven: done

src/feature_lib.cpp
zerofivefiveseven: done

include/feature_lib.h
zerofivefiveseven: done!

src/conversions.h
TimPushkin: You can remove `#include <opencv2/opencv.hpp>` here. I meant that you should have been replaced one with another.

examples/detection_example.cpp
```diff
@@ -0,0 +1,45 @@
...
+std::vector<uint8_t> matToVector(const cv::Mat &mat) {
+  std::vector<uint8_t> result;
+  cv::Mat cont = mat.isContinuous() ? mat : mat.clone();  // make sure the Mat is continuous
+  cont.reshape(1, 1).copyTo(result);  // copy from a 1D Mat
+  return result;
+}
+
+// Calculates keypoints and image descriptors
+// Takes an image file path as input
+// prints key points and image descriptors to stdout declared in associated header
```
TimPushkin: 1. Start this sentence with a capital letter.
2. `to stdout declared in associated header` -- what? You need to remove the info about the header, I believe.

include/feature_lib.h
```diff
@@ -0,0 +1,62 @@
...
+ public:
+  virtual DetectionResult detect(const std::vector<std::uint8_t> &input_vector) const = 0;
+
+  void setHeight(int height) {
+    mask.reshape(1, std::vector<int>{height, width});
+  }
+
+  void setWidth(int width) {
+    mask.reshape(1, std::vector<int>{height, width});
+  }
```
TimPushkin: You must also update the corresponding fields in both of these setters -- otherwise what's the point?

include/feature_lib.h
```diff
@@ -0,0 +1,62 @@
...
+class FeatureDetector {
+ public:
+  virtual DetectionResult detect(const std::vector<std::uint8_t> &input_vector) const = 0;
+
+  void setHeight(int height) {
+    mask.reshape(1, std::vector<int>{height, width});
+  }
+
+  void setWidth(int width) {
+    mask.reshape(1, std::vector<int>{height, width});
```
TimPushkin: I think, you cannot use `reshape` here (and for the height too) as the number of elements changes. You need to create a new matrix or change the size of this one somehow.

src/conversions.h
zerofivefiveseven: oh i am so sorry

zerofivefiveseven: done

include/feature_lib.h
```diff
@@ -0,0 +1,62 @@
...
+    return width;
+  }
+
+  int getHeight() const {
+    return height;
+  }
+
+ protected:
+  int width;
+  int height;
```
TimPushkin: You need to initialize them to something, or they'll stay uninitialized in case of using the default constructor. CLion gives you a warning -- do you even read them?

src/feature_lib.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include <opencv2/imgproc.hpp>
```
TimPushkin: Fix the order according to the code style

TimPushkin (changes requested):
You got the setters wrong again :(

examples/detection_example.cpp
zerofivefiveseven: my bad

src/feature_lib.cpp
zerofivefiveseven: image
https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

zerofivefiveseven: That's better?

include/feature_lib.h
zerofivefiveseven: Yep, this is because the names are the same as those in the setters
even i initialize them to something clion is still giving me warnings

zerofivefiveseven: when i change the parameter in the setter the warning goes away

zerofivefiveseven: maybe height_input and width_input ass parameters and also i initialize protected fields&

include/feature_lib.h
zerofivefiveseven: for example, when only the width increases and the height remains the same, you need to somehow add columns, so I will recreate it

include/feature_lib.h
TimPushkin: Yep, the warning I'm talking about appears in the default constructor's
 implementation.

Let the names be just `value` in the setters: `void setWidth(int value)`

src/feature_lib.cpp
TimPushkin: Yep, but remove that blank line

include/feature_lib.h
TimPushkin: Yep. There is also `cv::resize`, but I think that recreating would be faster because this resizing involves interpolation

src/feature_lib.cpp
```diff
@@ -0,0 +1,29 @@
...
+SiftDetector::SiftDetector() = default;
+
+SiftDetector::SiftDetector(int width, int height) {
+  this->width = width;
+  this->height = height;
+}
+
+featurelib::DetectionResult SiftDetector::detect(const std::vector<std::uint8_t> &input) const {
+  auto image = vectorToMat(input, height, width);
+  cv::cvtColor(image, image, cv::COLOR_RGB2GRAY);//
```
TimPushkin: Remove `//` in the end

TimPushkin (changes requested):
Other than this small detail, looks good to me

src/feature_lib.cpp
zerofivefiveseven: ok

examples/detection_example.cpp
```diff
@@ -0,0 +1,45 @@
+#include <iostream>
+#include <opencv2/core.hpp>
+#include <opencv2/imgcodecs.hpp>
+#include "../include/feature_lib.h"
```
achains: Is relative import really necessary here, since we have linked feature library to this target?

achains (commented):
Nice work, but it'd be better to remove relative import in examples

examples/detection_example.cpp
zerofivefiveseven: Nope, this works without a relative path as well. And the Google cpp codestyle says to avoid Unix ../ characters. I rewrite it like #include<"file.h">

TimPushkin (approved):

achains (approved):

include/feature_lib.h
```diff
@@ -0,0 +1,64 @@
...
+struct DetectionResult {
+  std::vector<KeyPoint> keypoints;
+  std::vector<std::vector<uint8_t>> descriptors;
+};
+
+class FeatureDetector {
+ public:
+  virtual DetectionResult detect(const std::vector<std::uint8_t> &input_vector) const = 0;
+
+  void setHeight(int value) {
```
anastasiia-kornilova: Подскажите по какому гайду мы живем, что методы определяются в хедере?

include/feature_lib.h
```diff
@@ -0,0 +1,64 @@
...
+    return height;
+  }
+
+ protected:
+  int width = 0;
+  int height = 0;
+  cv::Mat mask = cv::Mat(255 * cv::Mat::ones(height, width, CV_8U));
+};
+
+class SiftDetector : public FeatureDetector {
```
anastasiia-kornilova, line 44: Один класс на один файл или это тоже можно в нашем стайл гайде?

anastasiia-kornilova: Обсудили с Тимофеем -- так нужно скапиксу, пока отправляется в issues

include/feature_lib.h
```diff
@@ -0,0 +1,64 @@
+#ifndef FEATUREDETECT_FEATURELIB_H_
```
anastasiia-kornilova, line 1: Лучше обзывать всё Keypoints -- features слишком многозначное слово, можно пока занести в issue и поправить на след итерации

include/feature_lib.h
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  int getHeight() const {
+    return height;
+  }
+
+ protected:
+  int width = 0;
+  int height = 0;
+  cv::Mat mask = cv::Mat(255 * cv::Mat::ones(height, width, CV_8U));
```
anastasiia-kornilova: Мне неочевидно зачем нужна маска, давайте оставим коммент.

include/feature_lib.h
```diff
@@ -0,0 +1,64 @@
...
+  std::vector<std::vector<uint8_t>> descriptors;
+};
+
+class FeatureDetector {
+ public:
+  virtual DetectionResult detect(const std::vector<std::uint8_t> &input_vector) const = 0;
+
+  void setHeight(int value) {
+    height = value;
+    mask = cv::Mat(255 * cv::Mat::ones(height, width, CV_8U));
```
anastasiia-kornilova: Кажется этот конструктор здесь уместнее -- https://docs.opencv.org/4.x/d3/d63/classcv_1_1Mat.html#a3620c370690b5ca4d40c767be6fb4ceb

src/conversions.h
```diff
@@ -0,0 +1,36 @@
+#ifndef FEATUREDETECT_SRC_CONVERSIONS_H_
+#define FEATUREDETECT_SRC_CONVERSIONS_H_
+
+#include <opencv2/core.hpp>
+#include "feature_lib.h"
+
+cv::Mat vectorToMat(const std::vector<std::uint8_t> &input_vector, int height, int width) {
+  return {height, width, CV_8UC3, (std::uint8_t *) input_vector.data()};
```
anastasiia-kornilova: Лучше использовать явные касты, в данном случае static_cast. Почитать почему он умнее обычного explicit cast можно здесь https://cplusplus.com/doc/oldtutorial/typecasting/ .

src/conversions.h
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  return result;
+}
+
+std::vector<featurelib::KeyPoint> convertToStructure(const std::vector<cv::KeyPoint> &key_points) {
+  std::vector<featurelib::KeyPoint> output;
+  output.reserve(key_points.size());
+  for (auto &kp : key_points) {
+    featurelib::KeyPoint temp = {kp.pt.x, kp.pt.y, kp.angle, kp.response, kp.size};
```
anastasiia-kornilova: Ужаснее названия `temp` может быть только `a`, почему нельзя это в одну строчку сделать без временных переменных?

src/feature_lib.cpp
```diff
@@ -0,0 +1,29 @@
+#include <opencv2/imgproc.hpp>
+#include "feature_lib.h"
+#include "conversions.h"
+
+namespace featurelib {
+
+SiftDetector::SiftDetector() = default;
+
+SiftDetector::SiftDetector(int width, int height) {
+  this->width = width;
```
anastasiia-kornilova: Почитайте про то, что такое Initializer List in C++  и почему лучше так инициализировать параметры чем как у Вас

anastasiia-kornilova (changes requested):

src/conversions.h
zerofivefiveseven: Rewrote the method without using temporary variable
image
should i use some cast things there? i ll read link above rn

include/feature_lib.h
zerofivefiveseven: Done, should i use Initializer List there too?

anastasiia-kornilova: Not clear for me how do you plan to use it here, could you provide example?

src/conversions.h
anastasiia-kornilova: What would you like to cast to what? It is already the same type that is stored in vector.

include/feature_lib.h
zerofivefiveseven: image
made a comment in docstring style
should I use a multi-line comment on the left?

src/conversions.h
zerofivefiveseven: my bad, i messed up..

include/feature_lib.h
zerofivefiveseven: my bad, i messed up.. again

include/feature_lib.h
zerofivefiveseven: moved all base class definitions to .cpp file
declarations of all of this methods still in .h file

src/conversions.h
zerofivefiveseven: Static_cast from 'const unsigned char *' to 'std::uint8_t *' (aka 'unsigned char *') is not allowed 
Maybe const_cast instead?
const_cast
This type of casting manipulates the constness of an object, either to be set or to be removed. For example, in order to pass a const argument to a function that expects a non-constant parameter
if i use that type of casting it works

src/feature_lib.cpp
zerofivefiveseven: done.
I initialized the base class using this technique and in the initialization of the child class used that too
image
image

BuildOpenCV.cmake
```diff
@@ -26,7 +26,14 @@ if (NOT EXISTS ${OPENCV_DOWNLOAD_DIR}/opencv-${OPENCV_VERSION}/CMakeLists.txt)
                 ${OPENCV_DOWNLOAD_DIR}/${OPENCV_VERSION}.zip
                 SHOW_PROGRESS
         )
-        message(STATUS "Downloading OpenCV - done")
+        message(STATUS "Downloading OpenCV-extra")
+        file(
+                DOWNLOAD
+                https://github.com/opencv/opencv_contrib/archive/refs/heads/4.x.zip
```
TimPushkin: Donwload the same version as the version of the main package. You can do this from tags, for example: `https://github.com/opencv/opencv_contrib/archive/refs/tags/${OPENCV_VERSION}.zip`

BuildOpenCV.cmake
```diff
@@ -26,7 +26,14 @@ if (NOT EXISTS ${OPENCV_DOWNLOAD_DIR}/opencv-${OPENCV_VERSION}/CMakeLists.txt)
                 ${OPENCV_DOWNLOAD_DIR}/${OPENCV_VERSION}.zip
```
TimPushkin: Change it to `${OPENCV_DOWNLOAD_DIR}/opencv-${OPENCV_VERSION}.zip`. And don't forget to also change the `NOT EXISTS` check.

BuildOpenCV.cmake
```diff
@@ -26,7 +26,14 @@ if (NOT EXISTS ${OPENCV_DOWNLOAD_DIR}/opencv-${OPENCV_VERSION}/CMakeLists.txt)
...
                 SHOW_PROGRESS
         )
-        message(STATUS "Downloading OpenCV - done")
+        message(STATUS "Downloading OpenCV-extra")
+        file(
+                DOWNLOAD
+                https://github.com/opencv/opencv_contrib/archive/refs/heads/4.x.zip
+                ${OPENCV_DOWNLOAD_DIR}/opencv-extra.zip
```
TimPushkin: Change it to `${OPENCV_DOWNLOAD_DIR}/opencv-extra-${OPENCV_VERSION}.zip`

BuildOpenCV.cmake
```diff
@@ -26,7 +26,14 @@ if (NOT EXISTS ${OPENCV_DOWNLOAD_DIR}/opencv-${OPENCV_VERSION}/CMakeLists.txt)
                 ${OPENCV_DOWNLOAD_DIR}/${OPENCV_VERSION}.zip
                 SHOW_PROGRESS
         )
-        message(STATUS "Downloading OpenCV - done")
```
TimPushkin: Bring this back, please. And add a blank line after it.

BuildOpenCV.cmake
```diff
@@ -26,7 +26,14 @@ if (NOT EXISTS ${OPENCV_DOWNLOAD_DIR}/opencv-${OPENCV_VERSION}/CMakeLists.txt)
...
+        message(STATUS "Downloading OpenCV-extra")
```
TimPushkin: Add a separate check here that this hasn't yet been downloaded

BuildOpenCV.cmake
```diff
@@ -36,6 +43,14 @@ if (NOT EXISTS ${OPENCV_DOWNLOAD_DIR}/opencv-${OPENCV_VERSION}/CMakeLists.txt)
             DESTINATION ${OPENCV_DOWNLOAD_DIR}
     )
     message(STATUS "Extracting OpenCV - done")
+
+    message(STATUS "Extracting OpenCV-extra")
```
TimPushkin: Add a separate check that this hasn't yet been extracted (like the one I made for the standard OpenCV)

BuildOpenCV.cmake
```diff
@@ -65,6 +80,7 @@ set(
         OPENCV_CMAKE_ARGS  # https://docs.opencv.org/4.6.0/db/d05/tutorial_config_reference.html
         # General
         -DBUILD_LIST=${OPENCV_MODULES}
+        -DOPENCV_EXTRA_MODULES_PATH=${OPENCV_DOWNLOAD_DIR}/opencv-extra/opencv_contrib-4.x/modules
```
TimPushkin: Modify this path according to the changes I described above

TimPushkin (changes requested):
Better move the `opencv-contrib` downloading and extracting into a separate block with the same check that the standard `opencv` has. Maybe even create a function for it because the same code is used twice.

BuildOpenCV.cmake
TimPushkin: Also, add a `TODO` to specify only the extra modules we need. I believe, we only need `xfeatures2d`.

include/feature_lib.h
```diff
@@ -0,0 +1,56 @@
...
+  int getWidth() const;
+
+  int getHeight() const;
+
+ protected:
+  int width = 0;
+  int height = 0;
+  // You can limit your search to a set of regions with a mask.
+  // It must be a 8-bit integer matrix with non-zero values in the region of interest.
+  cv::Mat mask = cv::Mat(255 * cv::Mat::ones(height, width, CV_8U));
```
anastasiia-kornilova: Could you change here also to more suitable constructor as we discussed in other files?

src/conversions.h
anastasiia-kornilova: Yep, you are right

zerofivefiveseven:
I did something with commits, and somehow strangely rolled back the changes in BuildOpenCV.cmake

TimPushkin:
You did something wrong: you pushed the same commits and two lines related to the `opencv-contrib` changes are still there

zerofivefiveseven:
rolled back changes in BuildOpenCV.cmake

BuildOpenCV.cmake
TimPushkin: Closing because this and all the conversions below as this change has been removed

zerofivefiveseven:
rewrite history 
add last requested correction

anastasiia-kornilova (approved):

TimPushkin (approved):

PR #10, opened 2022-07-22
Menu for selection of the keypoint detection algorithm

alexdtat:
Implementation of keypoint detection algorithm's selection menu (Menu.kt). Here also was added enum class for algorithms (KeypointExtractionAlgorithm.kt) and map of string names of algorithms to their elements in enum.


alexdtat:
It seems that FeatureDetectAppType.kt is useless at this moment.

I am also not sure about `enum`-architecture (the key problem is that `enum`'s, it seems, are quite not suitable for being stored in `SharedPreferences`; they can be stored as indicies, but it looks uninformative imho; however, I also don't want to pass `String`'s into keypoint detection function).

alexdtat:
The current view of the menu GUI.
image

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -35,6 +40,7 @@ private const val RESOLUTION_HEIGHT = 360
 class MainActivity : ComponentActivity() {
     private val cameraExecutor = Executors.newSingleThreadExecutor()
     private val imageViewModel by viewModels<GrayscaleViewModel>()
+    private lateinit var currentAlgorithm: SharedPreferences
```
TimPushkin: It's wrong naming as it's not an algorithm -- it's just something where you retrive algorithm name from. Name it something like `preferences`.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -50,25 +56,33 @@ class MainActivity : ComponentActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
+        currentAlgorithm = this.getPreferences(Context.MODE_PRIVATE)
```
TimPushkin: You don't need `this` here

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -50,25 +56,33 @@ class MainActivity : ComponentActivity() {
...
+
         imageViewModel.isCameraPermissionGranted = cameraPermissionGranted()
         tryStartCamera()
 
         setContent {
+            val scaffoldState = rememberScaffoldState()
```
TimPushkin: Variables should be as deep as possible in the Compose structure so they can't be accessed from where they aren't needed. See the key point at the bottom of [this section](https://developer.android.com/jetpack/compose/state#state-hoisting).

And I also think that you shouldn't put anything outside of your theme as it won't have an effect this way.

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
```diff
@@ -0,0 +1,84 @@
...
+import androidx.compose.ui.Alignment.Companion.Bottom
+import androidx.compose.ui.Alignment.Companion.CenterVertically
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import androidx.core.content.edit
+import com.github.featuredetectandroid.utils.stringToAlgorithmMap
+
+private const val TAG = "Menu"
```
TimPushkin: Personally I believe that UI shouldn't produce any logs -- its one of the business logic's prerogatives

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
```diff
@@ -0,0 +1,84 @@
...
+
+@Composable
+fun Menu(currentAlgorithm: SharedPreferences) {
```
TimPushkin: Don't pass `SharedPreferences` here, a string plus a function that changes the value from the outside would be enough. Read about [state hoisting](https://developer.android.com/jetpack/compose/state#state-hoisting).

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
```diff
@@ -0,0 +1,84 @@
...
+    val radioOptions = stringToAlgorithmMap().keys.toList()
+    var selectedAlgorithm by remember {
+        mutableStateOf(currentAlgorithm.getString("algorithm", "None"))
+    }
+
+    Row(horizontalArrangement = Arrangement.Center) {
+        Text(
+            text = "Keypoint detection algorithm:",
+            modifier = Modifier.padding(all = 20.dp).align(Bottom),
```
TimPushkin: Place each method call on a separate line when there are more than one call. Android Studio does this for me when I press `Ctrl+Alt+L`.

TimPushkin: You can use `Row`'s `verticalAlignment` for aligning.

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
```diff
@@ -0,0 +1,84 @@
...
+            textAlign = TextAlign.Start,
+            fontSize = 20.sp
```
TimPushkin: Use typography file to set a `TextStyle` for this

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
```diff
@@ -0,0 +1,84 @@
...
+        )
+    }
+    radioOptions.forEach { text ->
+        Row(
+            Modifier.fillMaxWidth().selectable(
```
TimPushkin: Place it on separate lines

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
```diff
@@ -0,0 +1,84 @@
...
+                selected = (text == selectedAlgorithm),
+                onClick = {
```
TimPushkin: This should be replaced with a function provided as an argument

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
```diff
@@ -0,0 +1,84 @@
...
+                    currentAlgorithm.edit {
+                        putString("algorithm", text)
+                        apply()
+                    }
+                    selectedAlgorithm = text
+                }
+            )
+            Text(
+                text = text,
+                modifier = Modifier.align(CenterVertically),
```
TimPushkin: Use `Row`'s `verticalAlignment`

app/src/main/java/com/github/featuredetectandroid/utils/KeypointExtractionAlgorithm.kt
```diff
@@ -0,0 +1,17 @@
+package com.github.featuredetectandroid.utils
+
+enum class KeypointExtractionAlgorithm {
+    NONE,
+    SIFT,
+    SURF,
+    ORB,
+    SUPER_POINT
```
TimPushkin: SuperPoint is a single word

app/src/main/java/com/github/featuredetectandroid/utils/KeypointExtractionAlgorithm.kt
```diff
@@ -0,0 +1,17 @@
...
+}
+
+fun stringToAlgorithmMap() = mapOf(
```
TimPushkin: Try using `KeypointExtractionAlgorithm.valueOf()` for this

TimPushkin (changes requested):
Let's rename `FeatureDetectAppTheme.kt` and `FeatureDetectAppType.kt` to `Theme.kt` and `Type.kt` respectively as these are the default names everyone is used to

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
alexdtat: Fixed. Thanks for the link.

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
alexdtat: I deleted this logging call.

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/utils/KeypointExtractionAlgorithm.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/utils/KeypointExtractionAlgorithm.kt
alexdtat: I deleted `stringToAlgorithmMap()` for now.

UPD: it causes some troubles in Menu.kt.

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
TimPushkin: Put it back into the function that does the change as it's now in `MainActivity` -- it's great to log such kind of things

app/src/main/java/com/github/featuredetectandroid/utils/KeypointExtractionAlgorithm.kt
alexdtat: Now fixed.

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
alexdtat: fixed

alexdtat:
Perhaps it makes sense to add an uppercase cast to `getAlgorithmByName`. Although I am not sure.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -50,25 +57,38 @@ class MainActivity : ComponentActivity() {
...
-                Box(Modifier.fillMaxSize()) {
-                    if (imageViewModel.isCameraPermissionGranted) {
-                        imageViewModel.grayscaleBitmap?.let { bitmap ->
-                            Image(
-                                bitmap = bitmap.asImageBitmap(),
-                                modifier = Modifier.fillMaxSize(),
-                                contentDescription = "Grayscale photo"
+                val scaffoldState = rememberScaffoldState()
+                Scaffold(
+                    scaffoldState = scaffoldState,
```
TimPushkin: What do you need this for anyway? You don't use it, so just get rid of it.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -50,25 +57,38 @@ class MainActivity : ComponentActivity() {
...
+                    drawerContent = {
+                        Menu(
+                            preferences.getString("algorithm", "None") ?: "None",
```
TimPushkin: Let's create constants for this. Maybe even some `PreferencesHelper` class to hide this logic in -- if we add more preferences this may get cumbersome to maintain otherwise.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -133,4 +151,21 @@ class MainActivity : ComponentActivity() {
             }
         }, ContextCompat.getMainExecutor(this))
     }
+
+    private fun changeSelectedAlgorithmInPreferences(newAlgorithm: String) {
```
TimPushkin: Let's shorten this name. `selectAlgorithm` maybe?

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -133,4 +151,21 @@ class MainActivity : ComponentActivity() {
...
+
+        Log.i(
+            TAG,
+            "${
+            preferences.getString(
+                "algorithm",
+                "None"
+            )
+            } was selected. Algorithm name as selection argument: $newAlgorithm"
+        )
```
TimPushkin: Why is this log so long? Shouldn't `preferences.getString("algorithm", "None")` be equal to `newAlgorithm`? Even if not, you can still shorten it like so:
```kotlin
Log.i(
    TAG,
    "${preferences.getString("algorithm", "None")} was selected. " +
            "Algorithm name as selection argument: $newAlgorithm"
)
```
 Also, if you'll move the preferences logic into a separate class, you can also leave logging to it.

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
```diff
@@ -0,0 +1,77 @@
...
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Alignment.Companion.CenterVertically
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.unit.dp
+import com.github.featuredetectandroid.ui.theme.Typography
+import com.github.featuredetectandroid.utils.algorithmsNamesAsList
+
+@Composable
+fun Menu(currentAlgorithm: String, changeSelectedAlgorithmIntPreferences: (String) -> Unit) {
```
TimPushkin: Let's shorten `changeSelectedAlgorithmIntPreferences` here too

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
```diff
@@ -0,0 +1,77 @@
...
+    val radioOptions = algorithmsNamesAsList()
+    val (selectedAlgorithm, onAlgorithmSelection) = remember { mutableStateOf(currentAlgorithm) }
```
TimPushkin: I don't really get what's going on here, explain, please:
1. How `currentAlgorithm`, being a `String`, is destructured into two values.
2. What do we need `onAlgorithmSelection` for?

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    Row(horizontalArrangement = Arrangement.Center, verticalAlignment = Alignment.Bottom) {
```
TimPushkin: It's just a convention, but for me it seems like dividing this onto several lines would be more readable

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
```diff
@@ -0,0 +1,77 @@
...
+        Text(
+            text = "Keypoint detection algorithm:",
+            modifier = Modifier.padding(all = 20.dp),
+            textAlign = TextAlign.Start,
+            style = Typography.body1
+        )
+    }
+    radioOptions.forEach { text ->
```
TimPushkin: Let's add an empty line here

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
```diff
@@ -0,0 +1,77 @@
...
+            RadioButton(
+                selected = (text == selectedAlgorithm),
+                modifier = Modifier.padding(horizontal = 8.dp),
+                onClick = {
+                    onClickChangeSelectedAlgorithm(
+                        changeSelectedAlgorithmIntPreferences,
+                        onAlgorithmSelection,
+                        text
+                    )
+                }
```
TimPushkin: Do we really need this `onClick` if the whole `Row` is clickable? Why not just set it to `null`? You won't need this `onClickChangeSelectedAlgorithm` this way.

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
TimPushkin: You can probably name it something like `onAlgorithmSelected` after you apply the comments below

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
```diff
@@ -0,0 +1,77 @@
...
+fun Menu(currentAlgorithm: String, changeSelectedAlgorithmIntPreferences: (String) -> Unit) {
+    val radioOptions = algorithmsNamesAsList()
+    val (selectedAlgorithm, onAlgorithmSelection) = remember { mutableStateOf(currentAlgorithm) }
+
+    Row(horizontalArrangement = Arrangement.Center, verticalAlignment = Alignment.Bottom) {
+        Text(
+            text = "Keypoint detection algorithm:",
+            modifier = Modifier.padding(all = 20.dp),
+            textAlign = TextAlign.Start,
+            style = Typography.body1
```
TimPushkin: It should probably named `h6` (in the theme, I mean) as this is a relatively small header, not a body text. You can find examples [here](https://material.io/design/typography/the-type-system.html#type-scale).

TimPushkin: Also, don't use the elements from your theme like this -- it should be `MaterialTheme.typography.body1`. The whole point of the theme is that the UI doesn't depend on concrete theme classes, it just uses the values provided in the theme it is launched in.

app/src/main/java/com/github/featuredetectandroid/utils/KeypointExtractionAlgorithm.kt
```diff
@@ -0,0 +1,31 @@
+package com.github.featuredetectandroid.utils
+
+import android.util.Log
+
+private const val TAG = "KeypointExtractionAlgorithm"
+
+enum class KeypointExtractionAlgorithm {
```
TimPushkin: It's keypoint detection in our UI, not extraction -- let's be consistent

app/src/main/java/com/github/featuredetectandroid/utils/KeypointExtractionAlgorithm.kt
```diff
@@ -0,0 +1,31 @@
...
+    NONE,
+    SIFT,
+    SURF,
+    ORB,
+    SUPERPOINT
+}
+
+fun algorithmsNamesAsList() = enumValues<KeypointExtractionAlgorithm>().map {
```
TimPushkin: 1. It can be a simple `val`, no need to recalculate this each time.
2. Let's name it `names` and move it into the `companion object` of `KeypointExtractionAlgorithm` -- this would be more readable.
3. After applying the changes above, you can replace `enumValues<KeypointExtractionAlgorithm>()` with just `values()`. And also simplify the calls to `NONE` and `SUPERPOINT` below.

app/src/main/java/com/github/featuredetectandroid/utils/KeypointExtractionAlgorithm.kt
```diff
@@ -0,0 +1,31 @@
...
+    when (it) {
+        KeypointExtractionAlgorithm.NONE -> "None"
+        KeypointExtractionAlgorithm.SUPERPOINT -> "SuperPoint"
+        else -> it.name
+    }
+}
+
+fun getAlgorithmByName(name: String): KeypointExtractionAlgorithm {
```
TimPushkin: Seems like it's not used anywhere

app/src/main/java/com/github/featuredetectandroid/utils/KeypointExtractionAlgorithm.kt
```diff
@@ -0,0 +1,31 @@
...
+    var algorithm = KeypointExtractionAlgorithm.NONE
+    try {
+        algorithm = KeypointExtractionAlgorithm.valueOf(name)
+    } catch (illegalAlgorithmName: IllegalArgumentException) {
+        Log.e(TAG, "Illegal algorithm name: $name. $illegalAlgorithmName")
```
TimPushkin: This `TAG` is too long -- Android Studio gives a error

TimPushkin (changes requested):
Besides the notes below, let's rename `MainActivity.cameraPermissionGranted()` to `MainActivity.isCameraPermissionGranted()` -- makes it clearer

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
alexdtat: Yes, they _should_ be. So far, I have done this to display the correspondence of the result of the update to the argument.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/utils/KeypointExtractionAlgorithm.kt
alexdtat: > Seems like it's not used anywhere

This is the function to use to pass the selection of algorithm to the keypoint detection function (as `enum` element instead of `String`).

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
alexdtat: I used the third way to declare `MutableState`: https://developer.android.com/jetpack/compose/state#state-in-composables

`selectedAlgorithm` is the selected value and `onAlgorithmSelection` is the setter of it. I pass `onAlgorithmSelection` as an argument into `onClickChangeSelectedAlgorithm` to be able to change `MutableState`'s value. `currentAlgorithm` is algorithm by deafault (= selected at the moment of `Menu(...)` invokation).

alexdtat: But I can use something like that from their example if it'd be more readable: `onValueChange = { name = it }`

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
TimPushkin: I mean, may anything in `SharedPreferences`'s behavior go wrong, so we need to log it?

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
TimPushkin: Why you can't just do `val selectedAlgorithm by remember { mutableStateOf(currentAlgorithm) }` and set the value like `selectedAlgorithm = value`?

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
alexdtat: > I mean, does anything in `SharedPreferences`'s behavior may go wrong, so we need to log it?


Maybe something related to file processing. Although then an exception will probably be thrown. So it makes more sense to use the function argument as more quickly accessible information than the value from `SharedPreferences`.

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
alexdtat: > Why you can't just do `val selectedAlgorithm by remember { mutableStateOf(currentAlgorithm) }` and set the value like `selectedAlgorithm = value`?

I can do it. However, I don't know if this would be a more readable solution than passing the setter explicitly. Probably yes, though.

app/src/main/java/com/github/featuredetectandroid/utils/KeypointExtractionAlgorithm.kt
TimPushkin: Also, there is syntax for doing what you did here: `Log.e(TAG, "Illegal algorithm name: $name.", illegalAlgorithmName)`

TimPushkin: And don't give special names for errors in `catch` unless you need to -- let it be just `ex: IllegalArgumentException` or even `e: IllegalArgumentException`

app/src/main/java/com/github/featuredetectandroid/utils/KeypointExtractionAlgorithm.kt
TimPushkin: I think, you're using the word "stub" wrong here. Stub it something that does nothing and will be replaced later. And this is a function that you'll need to use later, do I understand you right?

alexdtat: > I think, you're using the word "stub" wrong here. Stub it something that does nothing and will be replaced later. And this is a function that you'll need to use later, do I understand you right?

Yes, I edited the comment. I meant the function that is unused for now, but is going to be used in the future.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
TimPushkin: Ok, if you believe we need it, leave it, otherwise, get rid of it

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
TimPushkin: For me it would be more readable if `onAlgorithmSelection` was named `setSelectedAlgorithm` as it's a setter.

Though, I believe you can get rid of `onClickChangeSelectedAlgorithm` all together if you remove `RadioButton`'s `onClick` like I said in other conversation.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
alexdtat: I removed this code after discussing the overall vision of the architecture.

app/src/main/java/com/github/featuredetectandroid/utils/KeypointExtractionAlgorithm.kt
alexdtat: I removed this function after discussing the overall vision of the architecture.

app/src/main/java/com/github/featuredetectandroid/utils/KeypointExtractionAlgorithm.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
alexdtat: Yes, because otherwise clicking on the button (namely, the round `RadioButton` itsef) will not have any effect, but clicking on the `Row` will. That is, as I understood from small experiments, clicking on a `Row` works on the entire area, except for the `RadioButton`, and clicking on the` RadioButton` works only for the `RadioButton` itself.

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
alexdtat: Fixed. About the `radioButton`, see below.

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/utils/KeypointExtractionAlgorithm.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
alexdtat: Decomposed, if I understood you correctly.

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
TimPushkin: When I set `RadioButton`'s `onClick` to `null` everything works as expected: the whole row is clickable including the button

TimPushkin: Padding needs to be adjusted though, because the button itself becomes unclickable and Compose removes the padding it adds to clickable elements

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
```diff
@@ -0,0 +1,84 @@
...
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Alignment.Companion.CenterVertically
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.unit.dp
+import com.github.featuredetectandroid.utils.KeypointDetectionAlgorithm
+
+@Composable
+fun Menu(currentAlgorithm: String, onAlgorithmSelected: (String) -> Unit) {
+    val radioOptions = KeypointDetectionAlgorithm.names
```
TimPushkin: Do we need this now when `KeypointDetectionAlgorithm.names` is a value? Actually, it may look more readable, so decide by yourself.

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
```diff
@@ -0,0 +1,84 @@
...
+    var selectedAlgorithm by remember { mutableStateOf(currentAlgorithm) }
```
TimPushkin: I think, it would be better to put this outside of this composable -- this better corresponds to the state hoisting pattern

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
```diff
@@ -0,0 +1,84 @@
...
+            Text(
+                text = text,
+                textAlign = TextAlign.Start,
+                style = MaterialTheme.typography.body1
+            )
+        }
+    }
+}
+
+fun onClickChangeSelectedAlgorithm(
```
TimPushkin: When `RadioButton`'s `onClick` is removed this can be removed too

app/src/main/java/com/github/featuredetectandroid/utils/PreferencesManager.kt
```diff
@@ -0,0 +1,25 @@
+package com.github.featuredetectandroid.utils
+
+import android.content.Context
+import android.util.Log
+import androidx.core.content.edit
+import com.github.featuredetectandroid.MainActivity
+
+private const val TAG = "PreferencesManager"
+private const val KEY = "algorithm"
+private const val NONE = "None"
```
TimPushkin: Let's call these `ALGORITHM_KEY` and `ALGORITHM_DEFAULT` as we may extend this class to handle more keys

app/src/main/java/com/github/featuredetectandroid/utils/PreferencesManager.kt
```diff
@@ -0,0 +1,25 @@
...
+
+class PreferencesManager(mainActivity: MainActivity) {
```
TimPushkin: Accept a type as general as possible. You don't need `MainActivity` here -- any `Activity` would be enough.

app/src/main/java/com/github/featuredetectandroid/utils/PreferencesManager.kt
```diff
@@ -0,0 +1,25 @@
...
+    private val preferences = mainActivity.getPreferences(Context.MODE_PRIVATE)
+
+    fun getAlgorithm() = preferences.getString(KEY, NONE) ?: NONE
```
TimPushkin: Let's name it `getSelectedAlgorithm`

app/src/main/java/com/github/featuredetectandroid/utils/PreferencesManager.kt
```diff
@@ -0,0 +1,25 @@
...
+
+    fun selectAlgorithm(newAlgorithm: String) {
```
TimPushkin: Let's name it `putSelectedAlgorithm` because if there'll be more keys figuring out special verbs like this for them will be a pain

TimPushkin (changes requested):
Some more changes needed

app/src/main/java/com/github/featuredetectandroid/utils/PreferencesManager.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
alexdtat: removed

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/utils/PreferencesManager.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/utils/PreferencesManager.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/utils/PreferencesManager.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/utils/PreferencesManager.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
alexdtat: Indeed, now we can inline it. But imho, for a person who is not used to work with `RadioButton`'s, it will be easier to understand the code when there is `radioOptions` value.

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -50,25 +59,38 @@ class MainActivity : ComponentActivity() {
...
-                                contentDescription = "Grayscale photo"
+                var selectedAlgorithm by remember {
+                    mutableStateOf(preferencesManager.getSelectedAlgorithm())
+                }
+                Scaffold(
+                    drawerContent = {
+                        Menu(selectedAlgorithm) { algorithmName ->
+                            preferencesManager.putSelectedAlgorithm(algorithmName)
+                            selectedAlgorithm = algorithmName
+                        }
```
TimPushkin: Let's use parameter names here as it's not obvious what this lambda represents

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
```diff
@@ -0,0 +1,61 @@
...
+    ) {
+        Text(
+            text = "Keypoint detection algorithm:",
+            modifier = Modifier.padding(all = 20.dp),
+            textAlign = TextAlign.Start,
+            style = MaterialTheme.typography.h6
+        )
+    }
+
+    radioOptions.forEach { text ->
```
TimPushkin: Let's use something more representative here instead of `text`. I'm not sure whta, though, as for the code below `algorithm` would suite best but then it's better to rename `radioOptions` to `algorithmNames` and then what's the point of creating this value if we can just inline `KeypointDetectionAlgorithm.names` as it has a similar meaning. The other option is to raname `text` to `option` but then it looks not so good in the code below.

To sum up, I'd go with `option` if you want `radioOptions` to stay.

TimPushkin: Or we can just pass the header text and radio options as parameters and rename everything to more general words so that this menu doesn't assume that it's used for selecting an algorithm.

If you do so, I'd suggest the following argument naming and order:
- `"Keypoint detection algorithm:"` -> `header`
- `radioOptions` -> `options`
- `selectedAlgorithm` -> `selectedOption`
- `onAlgorithmSelected` -> `onSelected`

app/src/main/java/com/github/featuredetectandroid/ui/theme/Type.kt
```diff
@@ -10,6 +10,6 @@ val Typography = Typography(
     body1 = TextStyle(
         fontFamily = FontFamily.Default,
         fontWeight = FontWeight.Normal,
-        fontSize = 16.sp
+        fontSize = 20.sp
```
TimPushkin: I think you need to set it back to the default value because `body1` is now as big as `h6` (which means options are as big as a header)

TimPushkin (changes requested):
A few small changes needed

app/src/main/java/com/github/featuredetectandroid/ui/theme/Type.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/ui/Menu.kt
alexdtat: > Or we can just pass the header text and radio options as parameters and rename everything to more general words so that this menu doesn't assume that it's used for selecting an algorithm.
> 
> If you do so, I'd suggest the following argument naming and order:
> 
> * `"Keypoint detection algorithm:"` -> `header`
> * `radioOptions` -> `options`
> * `selectedAlgorithm` -> `selectedOption`
> * `onAlgorithmSelected` -> `onSelected`

Good idea! Implemented.

TimPushkin (approved):

PR #13, opened 2022-07-23
SURF integration / opencv-contrib

zerofivefiveseven:
xfeatures2d which contains SURF appears in whitelist but don't in the library summary while build

### DIFF
--- a/BuildOpenCV.cmake
+++ b/BuildOpenCV.cmake
@@ -27,7 +27,6 @@ if (NOT EXISTS ${OPENCV_DOWNLOAD_DIR}/opencv-${OPENCV_VERSION}/CMakeLists.txt)
                 SHOW_PROGRESS
         )
         message(STATUS "Downloading OpenCV - done")
-
     endif ()
 
     message(STATUS "Extracting OpenCV")
@@ -54,7 +53,6 @@ if (NOT EXISTS ${OPENCV_DOWNLOAD_DIR}/opencv-contrib-${OPENCV_VERSION})
                 SHOW_PROGRESS
         )
         message(STATUS "Downloading OpenCV-contrib - done")
-
     endif ()
 
     message(STATUS "Extracting OpenCV-contrib")
--- a/BuildOpenCV.cmake
+++ b/BuildOpenCV.cmake
@@ -40,30 +40,29 @@ if (NOT EXISTS ${OPENCV_DOWNLOAD_DIR}/opencv-${OPENCV_VERSION}/CMakeLists.txt)
 endif ()
 
 # Download and extract OpenCV-contrib sources
-if (NOT EXISTS ${OPENCV_DOWNLOAD_DIR}/opencv-extra-${OPENCV_VERSION})
-    message(STATUS "OpenCV extra not extracted into ${OPENCV_DOWNLOAD_DIR}")
+if (NOT EXISTS ${OPENCV_DOWNLOAD_DIR}/opencv-contrib-${OPENCV_VERSION})
+    message(STATUS "OpenCV-contrib not extracted into ${OPENCV_DOWNLOAD_DIR}")
 
-    if (NOT EXISTS ${OPENCV_DOWNLOAD_DIR}/opencv-extra-${OPENCV_VERSION}.zip)
-        message(STATUS "OpenCV extra not downloaded into ${OPENCV_DOWNLOAD_DIR}")
+    if (NOT EXISTS ${OPENCV_DOWNLOAD_DIR}/opencv-contrib-${OPENCV_VERSION}.zip)
+        message(STATUS "OpenCV-contrib not downloaded into ${OPENCV_DOWNLOAD_DIR}")
 
-        message(STATUS "Downloading OpenCV extra")
+        message(STATUS "Downloading OpenCV-contrib")
         file(
                 DOWNLOAD
                 https://github.com/opencv/opencv_contrib/archive/refs/tags/${OPENCV_VERSION}.zip
-                ${OPENCV_DOWNLOAD_DIR}/opencv-extra-${OPENCV_VERSION}.zip
+                ${OPENCV_DOWNLOAD_DIR}/opencv-contrib-${OPENCV_VERSION}.zip
                 SHOW_PROGRESS
         )
-        message(STATUS "Downloading OpenCV extra - done")
-
+        message(STATUS "Downloading OpenCV-contrib - done")
     endif ()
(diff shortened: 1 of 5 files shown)


zerofivefiveseven:
yesterday it builds with xfeatures2d
working on it

zerofivefiveseven:
It builds, but xfeatures2d can't get in touch with project
[ 50%] Linking CXX shared library libfeatureDetect.so
/usr/bin/ld: cannot find -lopencv_xfeatures2d: No such file or directory

BuildOpenCV.cmake
```diff
@@ -16,28 +16,56 @@ endif ()
...
             ARCHIVE_EXTRACT
-            INPUT ${OPENCV_DOWNLOAD_DIR}/${OPENCV_VERSION}.zip
+            INPUT ${OPENCV_DOWNLOAD_DIR}/opencv-${OPENCV_VERSION}.zip
             DESTINATION ${OPENCV_DOWNLOAD_DIR}
     )
     message(STATUS "Extracting OpenCV - done")
 endif ()
 
+# Download and extract OpenCV-contrib sources
+if (NOT EXISTS ${OPENCV_DOWNLOAD_DIR}/opencv-extra-${OPENCV_VERSION})
```
TimPushkin: Let's replace `opencv-extra` here and `OpenCV extra` everywhere below with `opencv-contrib` and `OpenCV-contrib` respectively for consistency

BuildOpenCV.cmake
```diff
@@ -65,6 +93,8 @@ set(
         OPENCV_CMAKE_ARGS  # https://docs.opencv.org/4.6.0/db/d05/tutorial_config_reference.html
         # General
         -DBUILD_LIST=${OPENCV_MODULES}
+        -DOPENCV_EXTRA_MODULES_PATH=${OPENCV_DOWNLOAD_DIR}/opencv_contrib-${OPENCV_VERSION}/modules
+        -D OPENCV_ENABLE_NONFREE=ON
```
TimPushkin: Remove the space between `-D` and `OPENCV_ENABLE_NONFREE` for consistency

BuildOpenCV.cmake
```diff
@@ -16,28 +16,56 @@ endif ()
...
+    endif ()
+
+    message(STATUS "Extracting OpenCV extra")
+    file(
+            ARCHIVE_EXTRACT
+            INPUT ${OPENCV_DOWNLOAD_DIR}/opencv-extra-${OPENCV_VERSION}.zip
+            DESTINATION ${OPENCV_DOWNLOAD_DIR}
+    )
+    message(STATUS "Extracting OpenCV extra - done")
+endif ()
```
TimPushkin: Actually, exactly the same code is used here twice. You can create a function for this -- something like `function(get_opencv repo)` where `repo` argument will be either `opencv` or `opencv_contrib` (this one argument seems to be enough).

TimPushkin (changes requested):
Comments on CMake style.

C++ code will be reviewed by @anastasiia-kornilova later, I guess, but in general I'd say that:
1. You probably can move a lot of duplicated code from the detectors in their common abstract parent.
2. In the example it would be better to accept a command line argument with the algorithm name and run only the selected one because running all three is pretty messy.
3. Maybe we don't even need to print descriptors as it looks like a mess in the output -- just put descriptor size for each keypoint (in some algorithms the size may be dynamic, though, not in ours).

BuildOpenCV.cmake
```diff
@@ -16,28 +16,56 @@ endif ()
...
         message(STATUS "Downloading OpenCV")
         file(
                 DOWNLOAD
                 https://github.com/opencv/opencv/archive/refs/tags/${OPENCV_VERSION}.zip
-                ${OPENCV_DOWNLOAD_DIR}/${OPENCV_VERSION}.zip
+                ${OPENCV_DOWNLOAD_DIR}/opencv-${OPENCV_VERSION}.zip
                 SHOW_PROGRESS
         )
         message(STATUS "Downloading OpenCV - done")
+
```
anastasiia-kornilova: The base rule -- in PRs there should not be such useless changes. They usually appear when you do `git add` for the whole file. Read about command `git add -p` and start using it later to avoid such situations. That is classical professional git skills.

anastasiia-kornilova (approved):
I absolutely agree with @TimPushkin comments, but for now it looks like they are not crucial for MVP. I suggest to move all suggestions into separate issues and after the full ready working pipeline we will beautify those things.

BuildOpenCV.cmake
TimPushkin: Issue created

TimPushkin (approved):

BuildOpenCV.cmake
```diff
@@ -7,62 +7,45 @@ set(OPENCV_MODULES core,calib3d,imgproc,imgcodecs,features2d,highgui,xfeatures2d
 set(OPENCV_DOWNLOAD_DIR ${PROJECT_BUILD_DIR}/opencv-download)
 set(OPENCV_BUILD_DIR ${PROJECT_BUILD_DIR}/opencv-build)
 
+# Function which downloads and extracts OpenCV sources
+FUNCTION(GET_OPENCV REPO)
```
TimPushkin: Everything here should be lowercase

BuildOpenCV.cmake
```diff
@@ -7,62 +7,45 @@ set(OPENCV_MODULES core,calib3d,imgproc,imgcodecs,features2d,highgui,xfeatures2d
...
+
+        message(STATUS "Extracting ${REPO}")
+        file(
+                ARCHIVE_EXTRACT
+                INPUT ${OPENCV_DOWNLOAD_DIR}/${REPO}-${OPENCV_VERSION}.zip
+                DESTINATION ${OPENCV_DOWNLOAD_DIR}
+        )
+        message(STATUS "Extracting ${REPO} - done")
+    endif ()
+
```
TimPushkin: Remove this empty line

BuildOpenCV.cmake
```diff
@@ -7,62 +7,45 @@ set(OPENCV_MODULES core,calib3d,imgproc,imgcodecs,features2d,highgui,xfeatures2d
...
+ENDFUNCTION()
```
TimPushkin: This should be lowercase too

TimPushkin (changes requested):
Small style fixes required

TimPushkin (approved):

PR #21, opened 2022-07-25
SuperPoint integration and keypoints output.

alexdtat:
Implementation of keypoints', getting from `SuperPoint`, output. Selection of algorithm (`FetureDetector`) in `Menu` was added.


alexdtat:
At the moment I'm having trouble getting the image to center on the screen and the right size. I also noticed that by default the resolution of pictures from the camera is 640x480, and trying to set any (even this) value explicitly leads to a resolution of 720x720.

Also, I've noticed that setting a `Modifier.size()` on an `Image`, even with frame sizes, will itself scale the `Image` to fit the screen and center it. To see it, just add in `Image` invokation:
```
modifier = Modifier.size(
                width = imageViewModel.getSize().first.dp,
                height = imageViewModel.getSize().second.dp
            )
```

TimPushkin:
Haven't reviewed the whole code yet, but this code inside `Scaffold` gives me centered image with points drawn on top of it without any issues:
```kotlin
Box(
    modifier = Modifier.fillMaxSize(),
    contentAlignment = Alignment.Center
) {
    if (imageViewModel.isCameraPermissionGranted) {
        imageViewModel.grayscaleBitmap?.let { bitmap ->
            Image(
                bitmap = bitmap.asImageBitmap(),
                contentDescription = "Grayscale photo"
            )

            val (width, height) = with(LocalDensity.current) {
                bitmap.width.toDp() to bitmap.height.toDp()
            }

            Canvas(modifier = Modifier.size(width, height)) {
                drawPoints(
                    points = imageViewModel.keypointsOffset,
                    pointMode = PointMode.Points,
                    color = Color.Blue,
                    strokeWidth = 10f
                )
            }
        }
    } else {
        Text("Camera permission required")
    }
}
```

alexdtat:
Please, check commented `.setTargetResolution(...)` with different parameters on your device. I got different results on the emulator (Pixel 4 API 31) and on the physical device (Samsung Galaxy S10), since this option, as I understand it, is hardware dependent.

TimPushkin:
> Please, check commented `.setTargetResolution(...)` with different parameters on your device. I got different results on the emulator (Pixel 4 API 31) and on the physical device (Samsung Galaxy S10), since this option, as I understand it, is hardware dependent.

If I set it to `640x480` like it is in your commented code, I get something hardware-dependent. If I set it to `480 x 640`, I get `480 x 640` both on the emulator and on my Samsung Galaxy S10e. I figure, this is because our smartphones are in the portrait mode. so default for them is `480 x 640`, not `640x480` just like it's described in docs

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -14,35 +13,44 @@ import androidx.camera.core.CameraSelector
...
 import com.github.featuredetectandroid.utils.PhotoAnalyzer
 import com.github.featuredetectandroid.utils.PreferencesManager
+import com.github.featuredetectandroid.utils.selectFeatureDetector
 import java.util.concurrent.Executors
 
 private const val TAG = "MainActivity"
 
 private const val RESOLUTION_WIDTH = 640
-private const val RESOLUTION_HEIGHT = 360
+private const val RESOLUTION_HEIGHT = 480
```
TimPushkin: Let's just keep the default resolution for now and remove these

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -63,42 +71,17 @@ class MainActivity : ComponentActivity() {
         imageViewModel.isCameraPermissionGranted = isCameraPermissionGranted()
         tryStartCamera()
         preferencesManager = PreferencesManager(this)
+        preferencesManager.putSelectedAlgorithm(DEFAULT_ALGORITHM)
```
TimPushkin: Why do we need this? Let user keep their settings.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -143,9 +126,15 @@ class MainActivity : ComponentActivity() {
 
             val imageAnalyzer = ImageAnalysis.Builder()
                 .setOutputImageFormat(OUTPUT_IMAGE_FORMAT_YUV_420_888)
-                .setTargetResolution(Size(RESOLUTION_WIDTH, RESOLUTION_HEIGHT))
+                // .setTargetResolution(Size(RESOLUTION_WIDTH, RESOLUTION_HEIGHT))
+                // For some reason image is always 720 x 720 if this line is active.
```
TimPushkin: Let's just keep the standard resolution for now without using `setTargetResolution`

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -143,9 +126,15 @@ class MainActivity : ComponentActivity() {
...
                 .build()
-                .apply { setAnalyzer(cameraExecutor, PhotoAnalyzer(imageViewModel)) }
+                .apply {
+                    setAnalyzer(
+                        cameraExecutor,
+                        PhotoAnalyzer(imageViewModel)
+                    )
+                }
```
TimPushkin: Why have you decided to split this on several lines? It looked ok, but if some tool requires it to be split, let it be like this:
```kotlin
.apply {
    setAnalyzer(cameraExecutor, PhotoAnalyzer(imageViewModel))
}
```

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -159,4 +148,56 @@ class MainActivity : ComponentActivity() {
             }
         }, ContextCompat.getMainExecutor(this))
     }
+
+    @Composable
+    fun AppLayout(selectedAlgorithm: String, onAlgorithmSelected: (String) -> Unit) = Scaffold(
```
TimPushkin: Don't keep composables as methods. You can move it outside of `MainActivity`, but better create a separate file for it. Pass it what's in `onSelected` and `imageViewModel` as parameters.

TimPushkin: Also, let's move this `Scaffold` to the next line

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -159,4 +148,56 @@ class MainActivity : ComponentActivity() {
...
+                        this@MainActivity,
+                        algorithmName,
+                        imageViewModel.getSize().first,
+                        imageViewModel.getSize().second
+                    )
+                }
+            )
+        },
+        modifier = Modifier.fillMaxSize()
+    ) {
```
TimPushkin: `Scaffold` outputs `PaddingValues` here -- you must use it in your `Box`.

```kotlin
) { paddingValues ->
    Box(
        modifier = Modifier
            .fillMaxSize()
            .padding(paddingValues),
        ...
```
You can read about it in `Scaffold`'s docs (the description of `content` parameter).

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt
```diff
@@ -42,6 +44,17 @@ class PhotoAnalyzer(private val imageViewModel: GrayscaleViewModel) : ImageAnaly
             width = height.also { height = width }
         }
 
+        featureDetector?.width = width
+        featureDetector?.height = height
```
TimPushkin: Let's not change it for every frame as it doesn't really change. Determine it somehow from the camera after it's set and save somewhere.

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt
```diff
@@ -42,6 +44,17 @@ class PhotoAnalyzer(private val imageViewModel: GrayscaleViewModel) : ImageAnaly
...
+        val keypoints = (
+            featureDetector
+                ?.detect(luminanceArrayToRGB(oriented)) ?: Pair(
+                emptyList(),
+                emptyList()
+            )
+            )
+            .first
```
TimPushkin: Can you format it more compact somehow? Maybe like
```kotlin
val keypoints =
    (featureDetector?.detect(luminanceArrayToRGB(oriented))
        ?: Pair(emptyList(), emptyList())).first
```

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt
```diff
@@ -42,6 +44,17 @@ class PhotoAnalyzer(private val imageViewModel: GrayscaleViewModel) : ImageAnaly
...
+        imageViewModel.setKeypointsForOutput(keypoints)
         imageViewModel.setPicture(oriented, width, height)
```
TimPushkin: You probably don't need to pass sizes every time here too if you'll figure out how to retrieve them once

app/src/main/java/com/github/featuredetectandroid/utils/SelectFeatureDetector.kt
```diff
@@ -0,0 +1,15 @@
+package com.github.featuredetectandroid.utils
+
+import android.content.Context
+import com.github.featuredetectlib.learned.SuperPoint
+
+// TODO: add other FeatureDetector's after their implementation
+fun selectFeatureDetector(
```
TimPushkin: I think, this doesn't need its own file. We can place it in `MainActivity` as it won't be too big even with a lot of detectors.

app/src/main/java/com/github/featuredetectandroid/utils/SelectFeatureDetector.kt
```diff
@@ -0,0 +1,15 @@
...
+    context: Context,
+    algorithmName: String,
+    width: Int,
+    height: Int
+) = when (algorithmName) {
+    "SuperPoint" -> SuperPoint(context, width, height)
```
TimPushkin: We need to somehow tie `enum`s, names and classes in one place. Maybe create a class in which the `enum` will be nested and which will have methods like `enumToName` and `nameToClass`.

It's hard to maintain when these convertions are scattered all over the codebase.

app/src/main/java/com/github/featuredetectandroid/utils/SelectFeatureDetector.kt
TimPushkin: Actually, read the next comment

app/src/main/java/com/github/featuredetectandroid/utils/conversions/KeypointsToOffsetList.kt
```diff
@@ -0,0 +1,6 @@
+package com.github.featuredetectandroid.utils.conversions
+
+import androidx.compose.ui.geometry.Offset
+import com.github.featuredetectlib.Keypoint
+
+fun keypointsToOffsetList(keypoints: List<Keypoint>) = keypoints.map { Offset(it.x, it.y) }
```
TimPushkin: This function is so short that it could be inlined. Also it's used only in one place.

app/src/main/java/com/github/featuredetectandroid/utils/conversions/LuminanceArrayToRGB.kt
```diff
@@ -0,0 +1,7 @@
+package com.github.featuredetectandroid.utils.conversions
+
+private const val NUMBER_OF_COMPONENTS = 3
+
+fun luminanceArrayToRGB(luminanceByteArray: ByteArray) = ByteArray(
```
TimPushkin: Let's put this and `luminanceArrayToBitmap` into a single file and name it something like `LuminanceConversions`

app/src/main/java/com/github/featuredetectandroid/ui/OutputViewModel.kt
```diff
@@ -0,0 +1,33 @@
...
+import androidx.lifecycle.ViewModel
+import com.github.featuredetectandroid.utils.conversions.keypointsToOffsetList
+import com.github.featuredetectandroid.utils.conversions.luminanceArrayToBitmap
+import com.github.featuredetectlib.FeatureDetector
+import com.github.featuredetectlib.Keypoint
+
+class OutputViewModel : ViewModel() {
+    var grayscaleBitmap: Bitmap? by mutableStateOf(null)
+    var isCameraPermissionGranted by mutableStateOf(false)
+    var keypointsOffset: List<Offset> by mutableStateOf(emptyList())
```
TimPushkin: Let's name it `keypointOffsets` as it is a list of offsets

app/src/main/java/com/github/featuredetectandroid/ui/OutputViewModel.kt
```diff
@@ -0,0 +1,33 @@
...
+    private var width by mutableStateOf(0)
+    private var height by mutableStateOf(0)
+
+    fun setPicture(grayscaleByteArray: ByteArray, newWidth: Int, newHeight: Int) {
+        width = newWidth
+        height = newHeight
+        grayscaleBitmap = luminanceArrayToBitmap(grayscaleByteArray, newWidth, newHeight)
+    }
+
+    fun setKeypointsForOutput(keypoints: List<Keypoint>) {
```
TimPushkin: This is not needed as `keypointsOffset` is a `public var` with public setter anyway. Just set it using its setter.

app/src/main/java/com/github/featuredetectandroid/ui/OutputViewModel.kt
```diff
@@ -0,0 +1,33 @@
...
+        keypointsOffset = keypointsToOffsetList(keypoints)
+    }
+
+    fun getSize() = Pair(width, height)
```
TimPushkin: Its usage is pretty awkward. If you'll manage to set `width` and `height` only when they are really changed and not for each frame, it would be better to just make them `public` and set/get them from their setters/getters.

TimPushkin (changes requested):
Nothing crucial, but some changes are still needed

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
alexdtat: Now a `FeatureDetector` is set in the `onCreate` from `SharedPreferences` (previously this happened only when you clicked a button in the `Menu`).

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
alexdtat: Rudimentary code. Fixed.

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt
alexdtat: Current version is also a little bit strange, but it's ktlint's formatting result.

app/src/main/java/com/github/featuredetectandroid/ui/OutputViewModel.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/ui/OutputViewModel.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/utils/conversions/KeypointsToOffsetList.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/utils/conversions/LuminanceArrayToRGB.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt

app/src/main/java/com/github/featuredetectandroid/ui/OutputViewModel.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -60,52 +48,35 @@ class MainActivity : ComponentActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
-        imageViewModel.isCameraPermissionGranted = isCameraPermissionGranted()
+        outputViewModel.isCameraPermissionGranted = isCameraPermissionGranted()
         tryStartCamera()
         preferencesManager = PreferencesManager(this)
```
DmiitriyJarosh: can be lifted to the app level

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -115,7 +86,7 @@ class MainActivity : ComponentActivity() {
     }
```
DmiitriyJarosh, line 95: is it correct to call super before your call here?

app/src/main/java/com/github/featuredetectandroid/ui/AppLayout.kt
```diff
@@ -0,0 +1,81 @@
...
+    drawerContent = {
+        Menu(
+            header = "Keypoint detection algorithm:",
+            options = KeypointDetectionAlgorithm.names,
+            selectedOption = selectedAlgorithm,
+            onSelected = { algorithmName ->
+                preferencesManager.putSelectedAlgorithm(algorithmName)
+                onAlgorithmSelected(algorithmName)
+                outputViewModel.keypointOffsets = emptyList()
+                outputViewModel.featureDetector = selectFeatureDetector(
```
DmiitriyJarosh: why does viewmodel require context?

app/src/main/java/com/github/featuredetectandroid/utils/SelectFeatureDetector.kt
DmiitriyJarosh: possible solution is to store in enum value string with title and clazz with type, or builder lambda

alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/utils/SelectFeatureDetector.kt
alexdtat: deleted

app/src/main/java/com/github/featuredetectandroid/ui/AppLayout.kt
alexdtat: This is needed for the `SuperPoint` constructor, now I also handle this in `onDestroy` and `onResume`.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/ui/AppLayout.kt
TimPushkin: You don't need to do anything like this in `onDestroy` and `onResume` as `SuperPoint` doesn't hold any links to `Context` inside it. You can leave a comment where you pass the `Context` so that it won't be forgotten. I'll also add it to `SuperPoint`'s docs.

app/src/main/java/com/github/featuredetectandroid/FeatureDetectApp.kt
```diff
@@ -0,0 +1,19 @@
...
+    override fun onCreate() {
+        super.onCreate()
+        application = this
+        preferencesManager = PreferencesManager(applicationContext)
+    }
+
+    companion object {
+        private var application: FeatureDetectApp? = null
+        fun getPreferencesManager() = application?.preferencesManager
+    }
```
TimPushkin: You don't need all this. Just leave the `preferencesManager` property (and give it a `private set` as we don't want it to be changed from the outside) and access it like this in your `Activity`: `(application as FeatureDetectApp).preferencesManager`.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -14,36 +13,25 @@ import androidx.camera.core.CameraSelector
...
-private const val RESOLUTION_WIDTH = 640
-private const val RESOLUTION_HEIGHT = 360
-
 class MainActivity : ComponentActivity() {
     private val cameraExecutor = Executors.newSingleThreadExecutor()
-    private val imageViewModel by viewModels<GrayscaleViewModel>()
-    private lateinit var preferencesManager: PreferencesManager
+    private val outputViewModel by viewModels<OutputViewModel>()
+    private val preferencesManager = FeatureDetectApp.getPreferencesManager()
+        ?: error("FeatureDetectApp was not created!")
```
TimPushkin: This can be simplified as described above

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -60,62 +48,51 @@ class MainActivity : ComponentActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
-        imageViewModel.isCameraPermissionGranted = isCameraPermissionGranted()
+        outputViewModel.isCameraPermissionGranted = isCameraPermissionGranted()
         tryStartCamera()
```
TimPushkin: You can delete both of these lines as you have them in `onResume` which is called after `onCreate` so you just do the same work twice

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -60,62 +48,51 @@ class MainActivity : ComponentActivity() {
...
         super.onResume()
-        imageViewModel.isCameraPermissionGranted = isCameraPermissionGranted()
+        outputViewModel.isCameraPermissionGranted = isCameraPermissionGranted()
         tryStartCamera()
+        outputViewModel.featureDetector = KeypointDetectionAlgorithm.nameToFeatureDetector(
+            context = this,
+            algorithmName = preferencesManager.getSelectedAlgorithm(),
+            width = outputViewModel.width,
+            height = outputViewModel.height
+        )
```
TimPushkin: Why do you need this? As I said, `FeatureDetector` doesn't leak `Context`, but even if it did, you don't need it in `onResume` -- `onCreate` is enough.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -60,62 +48,51 @@ class MainActivity : ComponentActivity() {
...
     }
 
     override fun onDestroy() {
-        super.onDestroy()
         cameraExecutor.shutdown()
+        outputViewModel.featureDetector = null
```
TimPushkin: Redundant

app/src/main/java/com/github/featuredetectandroid/ui/AppLayout.kt
```diff
@@ -0,0 +1,80 @@
...
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.PointMode
+import androidx.compose.ui.graphics.asImageBitmap
+import androidx.compose.ui.platform.LocalDensity
+import com.github.featuredetectandroid.utils.KeypointDetectionAlgorithm
+import com.github.featuredetectandroid.utils.PreferencesManager
+
+@Composable
+fun AppLayout(
+    context: Context,
```
TimPushkin: You don't need to pass `Context` to composables, especially if you only need an `Activity` context -- you can get it from `LocalContext.current`

app/src/main/java/com/github/featuredetectandroid/ui/AppLayout.kt
```diff
@@ -0,0 +1,80 @@
...
+    outputViewModel: OutputViewModel,
+    preferencesManager: PreferencesManager,
+    selectedAlgorithm: String,
+    onAlgorithmSelected: (String) -> Unit
```
TimPushkin, line 28: You see, the majority of these arguments is only used inside `Menu`'s `onSelected`. Let's just pass what's you're passing there instead of `onAlgorithmSelected` (you can leave this naming, though) and get rid of the most of the arguments.

TimPushkin: After this, I think, we can also replace passing `ViewModel` here with passing its properties. This way we can remove `isCameraPermissionGranted` property from `ViewModel` as it'll only be used in `MainActivity`.

TimPushkin (changes requested):
I'll also try to make some changes regarding storing image `width` and `height` myself

app/src/main/java/com/github/featuredetectandroid/FeatureDetectApp.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
TimPushkin: Fixed this by myself

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
TimPushkin: Removed this by myself

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/ui/AppLayout.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/ui/AppLayout.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/ui/AppLayout.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -60,62 +48,57 @@ class MainActivity : ComponentActivity() {
...
         super.onCreate(savedInstanceState)
 
-        imageViewModel.isCameraPermissionGranted = isCameraPermissionGranted()
-        tryStartCamera()
-        preferencesManager = PreferencesManager(this)
+        preferencesManager = (application as FeatureDetectApp).preferencesManager
+
+        // SuperPoint, which needs a context during initialization,
+        // does not contain it inside itself, and therefore does not require any processing
+        // in the oDestroy and onResume.
```
TimPushkin: 1. Typo: `oDestroy`
2. Let's make it `nameToFeatureDetector`'s documentation.

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -60,62 +48,57 @@ class MainActivity : ComponentActivity() {
...
+                    keypointOffsets = outputViewModel.keypointOffsets,
+                    frameBitmap = outputViewModel.frameBitmap,
+                    selectedAlgorithm = selectedAlgorithm,
+                    onAlgorithmSelected = { algorithmName ->
+                        preferencesManager.putSelectedAlgorithm(algorithmName)
+                        selectedAlgorithm = algorithmName
+                        outputViewModel.keypointOffsets = emptyList()
+                        outputViewModel.featureDetector =
+                            KeypointDetectionAlgorithm.nameToFeatureDetector(
+                                algorithmName = preferencesManager.getSelectedAlgorithm(),
```
TimPushkin: Let's use `algorithmName` here -- should be quicker

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -60,62 +48,57 @@ class MainActivity : ComponentActivity() {
...
-                }
+                )
             }
         }
     }
 
     override fun onResume() {
         super.onResume()
-        imageViewModel.isCameraPermissionGranted = isCameraPermissionGranted()
         tryStartCamera()
```
TimPushkin, line 89: We always start the camera but never stop it in `onPause` (actually, I would put it in in `onStart` and `onStop` respectively). I see that it calls `unbindAll` before starting, so nothing bad happen, but let's create an issue about this and add later -- no need to listen to the camera when the activity isn't visible.

app/src/main/java/com/github/featuredetectandroid/ui/AppLayout.kt
```diff
@@ -0,0 +1,72 @@
...
+) {
+    Scaffold(
+        drawerContent = {
+            Menu(
+                header = "Keypoint detection algorithm:",
+                options = KeypointDetectionAlgorithm.names,
+                selectedOption = selectedAlgorithm,
+                onSelected = { algorithmName -> onAlgorithmSelected(algorithmName) }
```
TimPushkin: Can be simplified to `onSelected = onAlgorithmSelected`

TimPushkin (changes requested):
We're almost done here

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/ui/AppLayout.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/ui/AppLayout.kt
```diff
@@ -0,0 +1,72 @@
...
+    keypointOffsets: List<Offset>,
+    frameBitmap: Bitmap?,
+    selectedAlgorithm: String,
+    onAlgorithmSelected: (String) -> Unit
+) {
+    Scaffold(
+        drawerContent = {
+            Menu(
+                header = "Keypoint detection algorithm:",
+                options = KeypointDetectionAlgorithm.names,
```
DmiitriyJarosh, line 34: why not resources? what about translations?

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

DmiitriyJarosh (approved):

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
alexdtat: I created an issue for it.

app/src/main/java/com/github/featuredetectandroid/ui/AppLayout.kt
alexdtat: I created an issue for it.

TimPushkin (approved):

PR #30, opened 2022-07-26
Native integration (bridge)

alexdtat:
build.gradle module update for native library integration (bridge generation for JVM).


lib/src/main/cpp/include/surf_detector.h
```diff
@@ -0,0 +1,21 @@
...
+#include <opencv2/xfeatures2d.hpp>
+#include "feature_detector.h"
+
+namespace featurelib {
+
+class SurfDetector : public FeatureDetector {
+ public:
+  SurfDetector(int width, int height);
+
+  std::shared_ptr<DetectionResult> detect(const std::vector<std::uint8_t> &input) const override;
```
zerofivefiveseven, line 15: whether std::shared_ptr defined in header <memory>? If it's true, i can't find corresponding include anywhere

zerofivefiveseven (dismissed):

lib/src/main/cpp/include/surf_detector.h
TimPushkin: Nope, it's a transitive dependency: `#include <utility>` in `detection_result.h` -> `feature_detector.h` -> this file. That's not great, you're right.

lib/src/main/java/com/github/featuredetectlib/traditional/NativeKeypointWrapper.kt
```diff
@@ -0,0 +1,19 @@
...
+    override val x: Float
+        get() = keypoint.x
+    override val y: Float
+        get() = keypoint.y
+    override val strength: Float
+        get() = keypoint.strength
+    override val size: Float
+        get() = keypoint.size
+    override val angle: Float?
+        get() = keypoint.angle.takeIf { it != -1f } // angle == -1 if not applicable
```
alexdtat: Maybe we can make `-1f` a constant (or something like that)? Or there will be troubles with bridge code generation and typecasts?

TimPushkin: Good suggestion, everything'll be alright, will do

TimPushkin (approved):

PR #31, opened 2022-07-27
Native library solve issues  #15 #17

### LINKED ISSUE #15: Choose algorithm from CLI
In the example it would be better to accept a command line argument with the algorithm name and run only the selected one because running all three is pretty messy.

_Originally posted by @TimPushkin in https://github.com/TimPushkin/FeatureDetectAndroid/pull/13#pullrequestreview-1048599362_

### LINKED ISSUE #17: Descriptors output topic
Maybe we don't even need to print descriptors as it looks like a mess in the output -- just put descriptor size for each keypoint (in some algorithms the size may be dynamic, though, not in ours).

_Originally posted by @TimPushkin in https://github.com/TimPushkin/FeatureDetectAndroid/pull/13#pullrequestreview-1048599362_

zerofivefiveseven:
Now select the type of algorithm from the command line and output the length of the array of descriptors with key points

### DIFF
--- a/lib/src/main/cpp/BuildOpenCV.cmake
+++ b/lib/src/main/cpp/BuildOpenCV.cmake
@@ -2,7 +2,7 @@ set(PROJECT_BUILD_DIR ${CMAKE_SOURCE_DIR}/build)
 
 set(OPENCV_VERSION 4.6.0)
 # TODO: remove unused modules
-set(OPENCV_MODULES core,calib3d,imgproc,imgcodecs,features2d,highgui,xfeatures2d)
+set(OPENCV_MODULES core,calib3d,imgproc,imgcodecs,features2d,xfeatures2d)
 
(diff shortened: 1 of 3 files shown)


TimPushkin (changes requested):
Seems too complex.

To iterate through two vectors at the same time you can just do something like this as we expect them to have the same length (pseudocode):
```kotlin
for (i in 0..(keypoints.size() - 1)) {
    // do something with keypoints[i] and descriptors[i]
}
```

For selecting an algorithm let's just go with this

```c
if (name == "SIFT") // run SIFT
else if (name == "SURF") // run SURF
...
```
This is because we want to keep examples as easy to understand as possible.

TimPushkin:
Solves #15 and #17

TimPushkin:
Solves #15
Solves #17

TimPushkin:
Superseded by #66

PR #32, opened 2022-07-27
Remove unused OpenCV modules

zerofivefiveseven:
removed unused modules in buildopencv.cmake
add imgdecodes(imread) linkage to example and remove that dependency from lib proj


TimPushkin (approved):
Linking OpenCV to the example like you did didn't really work. This is because the values found in the file linking OpenCV are not available in the file that defines examples -- their scope is limited to a file.

PR #33, opened 2022-07-27
Time measurement for algorithms' performance check

alexdtat:
Average [edited: Latest] keypoint detection time measurement and output were added.


alexdtat:
The strange workaround with `isDetectorChanged` is due to the fact that in a naive implementation, when the algorithm changes, the last received frame would continue processing with the old `FeatureDetector`, but would write the data into the statistics of the new one. This can be critical when moving between algorithms with very different performance. In the future, this can be corrected, for example, in issues.

TimPushkin:
Why not just set frames and time to 0 when detector changes?

TimPushkin (commented):
Let's just display the last detection time in this PR and make a button to calculate mean time in another branch which we won't merge in `dev`

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt
```diff
@@ -78,6 +79,7 @@ class MainActivity : ComponentActivity() {
                                 width = outputViewModel.featureDetector?.width ?: 0,
                                 height = outputViewModel.featureDetector?.height ?: 0
                             )
+                        outputViewModel.milliseconds = 0
```
TimPushkin: It's not a problem if calculation time of the last frame is still on screen even if the detector has changed, so let's remove this

app/src/main/java/com/github/featuredetectandroid/ui/AppLayout.kt
```diff
@@ -37,20 +43,21 @@ fun AppLayout(
...
         modifier = Modifier.fillMaxSize()
     ) { paddingValues ->
-        Box(
+        Column(
             modifier = Modifier
                 .fillMaxSize()
                 .padding(paddingValues),
-            contentAlignment = Alignment.Center
+            verticalArrangement = Arrangement.Center,
+            horizontalAlignment = CenterHorizontally
```
TimPushkin: Change this to `Alignment.CenterHorizontally`

app/src/main/java/com/github/featuredetectandroid/ui/AppLayout.kt
```diff
@@ -63,8 +70,21 @@ fun AppLayout(
                 Image(
                     bitmap = bitmap,
                     contentDescription = "Grayscale photo",
-                    modifier = Modifier.fillMaxSize()
+                    modifier = Modifier
+                        .fillMaxSize()
+                        .weight(1f)
                 )
+
+                if (keypointOffsets.isNotEmpty()) {
```
TimPushkin: Why not `selectedAlgorithm != KeypointDetectionAlgorithm.NONE.algorithmName`?

app/src/main/java/com/github/featuredetectandroid/ui/AppLayout.kt
```diff
@@ -63,8 +70,21 @@ fun AppLayout(
...
+                    Text(
+                        text = "Last detection time: $milliseconds ms.",
```
TimPushkin: I think, `Detection time` is enough -- it's pretty obvious that it's the most recent value as it updates frequently

app/src/main/java/com/github/featuredetectandroid/ui/AppLayout.kt
```diff
@@ -63,8 +70,21 @@ fun AppLayout(
...
+                        modifier = Modifier.padding(
+                            vertical = 30.dp,
+                            horizontal = 20.dp
+                        ),
+                        style = MaterialTheme.typography.body1
```
TimPushkin: I think, it is the default in this case

TimPushkin: You can also remove this in `Menu` (59th line)

app/src/main/java/com/github/featuredetectandroid/ui/OutputViewModel.kt
```diff
@@ -15,4 +15,5 @@ class OutputViewModel : ViewModel() {
     var frameBitmap: Bitmap? by mutableStateOf(null)
     var keypointOffsets: List<Offset> by mutableStateOf(emptyList())
     var featureDetector: FeatureDetector? by mutableStateOf(null)
+    var milliseconds: Long by mutableStateOf(0)
```
TimPushkin: Let's call it `calcTimeMs` for instance, because it's not obvious what it represents

app/src/main/java/com/github/featuredetectandroid/ui/AppLayout.kt
```diff
@@ -2,27 +2,33 @@ package com.github.featuredetectandroid.ui
...
 import com.github.featuredetectandroid.utils.KeypointDetectionAlgorithm
 
+private const val KEYPOINT_WIDTH = 10f
+
 @Composable
 fun AppLayout(
     isCameraPermissionGranted: Boolean,
     keypointOffsets: List<Offset>,
     frameBitmap: Bitmap?,
+    milliseconds: Long,
```
TimPushkin: Rename this the same way as the corresponding `ViewModel` property

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt
```diff
@@ -46,7 +47,11 @@ class PhotoAnalyzer(private val outputViewModel: OutputViewModel) : ImageAnalysi
             outputViewModel.featureDetector?.let { detector ->
                 if (detector.width != width) detector.width = width
                 if (detector.height != height) detector.height = height
+                val startTime = SystemClock.elapsedRealtime()
                 val (keypoints, _) = detector.detect(luminanceArrayToRGB(oriented))
+                if (outputViewModel.keypointOffsets.isNotEmpty()) {
```
TimPushkin: We want to show time if `outputViewModel.featureDetector != null`, so this check is redundant

TimPushkin (changes requested):
Needs a few changes

app/src/main/java/com/github/featuredetectandroid/MainActivity.kt

app/src/main/java/com/github/featuredetectandroid/ui/AppLayout.kt
alexdtat: > Change this to `Alignment.CenterHorizontally`

Android Studio says: "Redundant qualifier name". But I can add this for more explicitness.

app/src/main/java/com/github/featuredetectandroid/ui/AppLayout.kt
TimPushkin: Ok, it's probably better to leave this as it is because even if the algo has changed the keypoints may still be present (we set them to `emptyList` when this happens but still)

app/src/main/java/com/github/featuredetectandroid/ui/AppLayout.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt

TimPushkin (changes requested):
I also think that we don't need `outputViewModel.keypointOffsets = emptyList()` when algo changes because keypoints from the previous algo may still appear if they are currently being calculated. So it's useless.

app/src/main/java/com/github/featuredetectandroid/ui/AppLayout.kt
TimPushkin: This is because you have `import androidx.compose.ui.Alignment.Companion.CenterHorizontally` -- change it to `import androidx.compose.ui.Alignment`

alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/ui/OutputViewModel.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/ui/AppLayout.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/ui/AppLayout.kt
alexdtat: fixed

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt
```diff
@@ -46,7 +47,9 @@ class PhotoAnalyzer(private val outputViewModel: OutputViewModel) : ImageAnalysi
             outputViewModel.featureDetector?.let { detector ->
                 if (detector.width != width) detector.width = width
                 if (detector.height != height) detector.height = height
+                val startTime = SystemClock.elapsedRealtime()
                 val (keypoints, _) = detector.detect(luminanceArrayToRGB(oriented))
+                outputViewModel.calcTimeMs = SystemClock.elapsedRealtime() - startTime
```
TimPushkin, line 53: So, here we also measure the time `luminanceArrayToRGB` take which we don't want to. Save it to `rgbArray` and pass it as an argument.

TimPushkin (changes requested):
Noticed another important issue

app/src/main/java/com/github/featuredetectandroid/utils/PhotoAnalyzer.kt
alexdtat: fixed

TimPushkin (approved):

PR #35, opened 2022-07-28
Make image fullscreen

### LINKED ISSUE #24: Add output image (with keypoints) scaling
Now, for small camera resolutions, an image of the same size is displayed, but it is preferable to have an image with keypoints scaled according to the screen parameters.

TimPushkin:
Now output image dynamically occupies the whole screen.

**Note**: the bitmap is required to be mutable.


alexdtat (changes requested):
The application has become much better! I suggest only adding information about the `mutable Bitmap` as a KDoc comment to the `ViewModel`, for example.

alexdtat (approved):
We're done here. :)

PR #37, opened 2022-07-28
RGB colorful image output instead of old grayscale.

### LINKED ISSUE #22: Add RGB support instead of YUV 420 888 luminance plane (grayscale image)
Libraries receive an RGB image as input and convert them to the required format themselves, so there is no need for restrictions on the output of a grayscale image.

alexdtat:
RGB colorful image output instead of old grayscale.


alexdtat:
So far, I have left the code for the grayscale conversions.

alexdtat:
App's appearance at the moment:

Screenshot_20200129-042048_FeatureDetectAndroid 1

alexdtat:
And another image:

Screenshot_20200129-042611_FeatureDetectAndroid 1

TimPushkin (commented):
You created hell of a naming, especially in `rgbaComponentsToBitmap`.

It took me pretty long to understand why `rgbaComponentsToBitmap` did this weird indexing because you had wrong naming in your color channels. It turned out that a `Bitmap.Config.ARGB_8888` pixel is actually stored in ABGR channel order.

So, I did a lot of renaming and rearrangements, especially in `rgbaComponentsToBitmap` and added some docs. Please, take a look.

alexdtat:
Thank you very much! Now code is much cleaner and easier to understand.

TimPushkin (approved):

PR #38, opened 2022-07-29
Configure release build

TimPushkin:
- Configures `release` build variant
- Removes unused test dependecnies of `app`
- Updates Android Compose and Kotlin

`lib`'s ProGuard rules are not optimal yet.


alexdtat (approved):
Well done!

PR #41, opened 2022-07-29
Fix startup-without-permissions release crash

TimPushkin:
Without this fix app crashes on release build when starting without camera permissions. Seems like an Android Compose bug.


alexdtat (approved):
Thanks! It helped.

Issue #53, opened 2022-08-07
Update screenshots in README


TimPushkin:
Currently the main `README` has outdated GUI screenshots which need to be updated

TimPushkin:
I'm planning to do this after dealing with resolution selection

PR #60, opened 2022-08-13
Add file analysis

TimPushkin:
Implements image analysis from a file with logging


app/src/main/java/com/github/kpdandroid/utils/detection/DetectionLogger.kt
```diff
@@ -0,0 +1,41 @@
...
+import java.io.OutputStream
+
+private const val TAG = "DetectionLogger"
+
+class DetectionLogger(logFile: OutputStream) : Closeable {
+    private val writer = logFile.bufferedWriter()
+
+    fun log(detectorName: String?, width: Int, height: Int, keypointsNum: Int, timeMs: Long) {
+        try {
+            writer.write("$detectorName ${width}x$height $keypointsNum $timeMs")
```
alexdtat, line 15: Maybe it'd be better to make braces consistent?

app/src/main/java/com/github/kpdandroid/utils/detection/DetectionRunners.kt
```diff
@@ -0,0 +1,81 @@
...
+
+        // Combined mean: m = ((n1 * m1) + (n2 * m2)) / (n1 + n2)
+        //
+        // Where:
+        // n1, n2 -- numbers of elements in each set
+        // m1, m2 -- means of each set
+        mean = (n * oldMean + time) / (n + 1)
+
+        // Combined unbiased estimation of standard deviation:
+        // https://math.stackexchange.com/a/2971563/1085312
```
alexdtat, line 61: Wouldn't we like to give the user this link in the documentation?

app/src/main/java/com/github/kpdandroid/utils/camera/SnapshotAnalyzer.kt
```diff
@@ -0,0 +1,54 @@
...
+) : ImageAnalysis.Analyzer {
+    override fun analyze(image: ImageProxy) {
+        val width = image.width
+        val height = image.height
+        val (snapshot, rowStride, pixelStride) =
+            image.planes[0].run { Triple(buffer.toByteArray(), rowStride, pixelStride) }
+        Log.v(
+            TAG,
+            "Analyzing snapshot:\n" +
+                "- size: ${width}x$height\n" +
```
alexdtat, line 28: Maybe it'd be better to make braces consistent?

app/src/main/java/com/github/kpdandroid/utils/camera/SnapshotAnalyzer.kt
```diff
@@ -0,0 +1,54 @@
...
+                "- pixel stride: $pixelStride\n" +
+                "- row stride: $rowStride\n" +
+                "- width * pixelStride: ${width * pixelStride}"
+        )
+
+        val (keypoints, calcTimeMs) = detectorGetter()?.detectTimed(
+            rgbBytes = rgbaBytesToRgbBytes(snapshot, width, height, rowStride, pixelStride),
+            imageWidth = width,
+            imageHeight = height
+        ) ?: (emptyList<Offset>() to 0L)
```
alexdtat: Wouldn't a constant be better here instead of `0L`?

app/src/main/java/com/github/kpdandroid/utils/detection/DetectionAlgo.kt
```diff
@@ -0,0 +1,42 @@
...
+    companion object {
+        val titles = values().map { it.title }
+
+        fun from(detector: KeypointDetector?) = when (detector) {
+            is Sift -> SIFT
+            is Surf -> SURF
+            is Orb -> ORB
+            is SuperPoint -> SUPERPOINT
+            null -> NONE
+            else -> null
```
alexdtat, line 26: So we prefer to handle nullable values? Compared to `NONE` in `else` will it be better because of abscence of errors' hiding?

app/src/main/java/com/github/kpdandroid/utils/PreferencesManager.kt
```diff
@@ -3,38 +3,59 @@ package com.github.kpdandroid.utils
...
-            Log.i(TAG, "Algorithm $value has been selected.")
+            prefs.edit { putString(CAMERA_ALGO_KEY, value) }
+            cameraAlgoNameState = value
+            Log.i(TAG, "Algorithm $value has been selected for camera analysis.")
         }
 
-    var selectedResolution: Size?
-        get() = preferences.getString(RESOLUTION_KEY, RESOLUTION_DEFAULT)
-            ?.let { Size.parseSize(it) }
+    private var resolutionState by mutableStateOf(
```
alexdtat: There seems to be a lot of boilerplate code here, but I don't have much idea how to fix it in a concise and maintainable way. If you suddenly have any ideas, you can create an issue.

alexdtat (changes requested):
Unfortunately, I had problems with the hardware, so I just looked at the code on the github. Overall looks great, but there are a few stylistic issues. In addition, somewhere we use a `name` for algorithms, and somewhere a `title`. Are we satisfied with such synonyms or do we want uniformity? I'll try to look through the rest of the files in the morning.

app/src/main/java/com/github/kpdandroid/utils/detection/DetectionLogger.kt
TimPushkin: Braces are used for `width` because without them it would be parsed as `widthx`. No need to use braces for other variables here.

app/src/main/java/com/github/kpdandroid/utils/detection/DetectionRunners.kt
TimPushkin: This comment is intended for code readers to better understand the formula below. I think, stating "standard deviation" in the docs and UI is enough. Users don't really have to know how it's calculated: docs are about what a function does, not how it does it.

app/src/main/java/com/github/kpdandroid/utils/camera/SnapshotAnalyzer.kt
TimPushkin: Again, this is because without them we'll get `widthx`. I doubt that adding braces to other variables will improve readability: seems like it'll make the string more cumbersome to read.

app/src/main/java/com/github/kpdandroid/utils/camera/SnapshotAnalyzer.kt
TimPushkin: Maybe it's better to use `null` here, so it doesn't get displayed in UI at all, I'll see

app/src/main/java/com/github/kpdandroid/utils/detection/DetectionAlgo.kt
TimPushkin: Yep, I figured this'll be better as it doesn't add any boilerplate code, as far as I remember, and makes the function more straightforward

app/src/main/java/com/github/kpdandroid/utils/PreferencesManager.kt
TimPushkin: I wrote this code precisely because I have no idea how to make it better :)

I would like to replace `State`s with `Flow`s, so that `PreferencesManager` wouldn't be so tied to UI, but it'll only create more code.

app/src/main/java/com/github/kpdandroid/utils/detection/DetectionLogger.kt
TimPushkin: It's a debatable subject, but let's settle it like this in this project: we use braces only where they are required. Firstly, this makes code more compact, secondly, I believe, by default IntelliJ IDEA-based IDEs will even give you a warning if you use braces where unnecessary.

TimPushkin:
Speaking about `title` and `name`, yeah, I don't like this either and thought a little about this. I think, I'll go with `title` everywhere as it doesn't collide with `enum`'s standard property name and is pretty short.

Upd: done.

app/src/main/java/com/github/kpdandroid/utils/PreferencesManager.kt
alexdtat: Ok, I get it.

app/src/main/java/com/github/kpdandroid/utils/camera/SnapshotAnalyzer.kt
TimPushkin: Replaced with `null` as it already has a corresponding message in UI exactly for this case

app/src/main/java/com/github/kpdandroid/ui/BottomMenu.kt
```diff
@@ -18,26 +23,100 @@ import androidx.compose.runtime.remember
...
+    val onSelected: (String) -> Unit
+)
+
+@OptIn(ExperimentalMaterialApi::class)
+@Composable
+fun ExpandableBottomMenuItem(
+    content: ExpandableBottomMenuItemContent,
+    leadingIcon: (@Composable BoxScope.() -> Unit)? = null
 ) {
     var expanded by remember { mutableStateOf(false) }
```
alexdtat, line 121: Wouldn't it be better to rename it to `isExpanded` or something like that to emphasize `Boolean`?

app/src/main/java/com/github/kpdandroid/MainActivity.kt
```diff
@@ -4,152 +4,151 @@ import android.Manifest
...
+    private fun isPermissionGranted(permission: String) =
+        ContextCompat.checkSelfPermission(this, permission) == PackageManager.PERMISSION_GRANTED
 
-    private fun shouldRationalize() =
+    private fun shouldRationalize(permission: String) =
         (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.M) &&
-            shouldShowRequestPermissionRationale(Manifest.permission.CAMERA)
+            shouldShowRequestPermissionRationale(permission)
+
+    // Not using onResume to account for changes made in split screen
```
alexdtat, line 148: Do we want to add this to the documentation? Or is the information only for working with the code?

app/src/main/java/com/github/kpdandroid/ui/DetectionResultColumn.kt
```diff
@@ -0,0 +1,56 @@
...
+import androidx.compose.material.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.ImageBitmap
+import androidx.compose.ui.unit.dp
+
+@Composable
+fun DetectionResultColumn(
+    imageLayers: List<ImageBitmap>,
```
alexdtat, line 18: Why do we use a `Pair` (ImagesAnalysisViewModel.kt) for layers somewhere, and a `List` somewhere?

app/src/main/java/com/github/kpdandroid/ui/viewmodels/FileAnalysisViewModel.kt
```diff
@@ -0,0 +1,125 @@
...
+            return
+        }
+
+        viewModelScope.launch(Dispatchers.Default) {
+            if (!detectionRunningJob.run { isCompleted || isCancelled }) stopDetection()
+
+            Log.i(TAG, "Running detection for $times times.")
+
+            detectionRunningJob = launch {
+                detectionProgress = 0.0f
```
alexdtat, line 63: Let's maybe add a constant here (possibly with an indication that this is a share). And do we want to use `Float` here or would it be enough, for example, `Int` percentages?

alexdtat: Or we can add `Share` in the variable's name.

app/src/main/java/com/github/kpdandroid/ui/screens/FileAnalysisScreen.kt
```diff
@@ -0,0 +1,231 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+
+@Composable
+// - Not using `vm = viewModel()` as we need to use exactly the same instance as the MainActivity
+// does.
+// - Not using `vm = viewModel(LocalContext.current as ViewModelStoreOwner)` as it may cause
+// unexpected cast exceptions.
+fun FileAnalysisScreen(vm: FileAnalysisViewModel) {
+    var openDialog by rememberSaveable { mutableStateOf(false) }
```
alexdtat, line 52: Maybe `isDialogOpen` would be better (`openDialog` looks like some action/command).

app/src/main/java/com/github/kpdandroid/ui/screens/FileAnalysisScreen.kt
```diff
@@ -0,0 +1,231 @@
...
+                    "Mean detection time: ${formatTime(calcTimeMs)} ms.",
+                    "Standard deviation: ${formatTime(stdDevMs)} ms."
+                )
+            } ?: listOf("Analyze an image to see detection time"),
+            modifier = Modifier.padding(paddingValues)
+        )
+    }
+}
+
+private fun formatTime(time: Double) = "%5f".format(time)
```
alexdtat, line 101: Here, too, we could use a constant for easier readability and editability of the code, although for a function in 1 line this is already insignificant (perhaps even harmful). So the decision is up to you.

app/src/main/java/com/github/kpdandroid/ui/screens/FileAnalysisScreen.kt
```diff
@@ -0,0 +1,231 @@
...
+
+private const val MIN_RUN_TIMES = 1
+
+@Composable
+private fun RunConfigurationDialog(open: Boolean, onDismiss: () -> Unit, onConfirm: (Int) -> Unit) {
```
alexdtat: The same question about `Boolean` naming.

app/src/main/java/com/github/kpdandroid/ui/screens/FileAnalysisScreen.kt
```diff
@@ -0,0 +1,231 @@
...
+                    "Pick log" to { logPicker.launch(SUGGESTED_LOG_FILENAME) }
+                ),
+            )
+        }
+    )
+}
+
+@Composable
+private fun FileAnalysisFab(
+    show: Boolean,
```
alexdtat: The same question about Boolean naming.

alexdtat (changes requested):
At the moment, I have looked through the entire code, but for testing the functionality (and maybe reviewing the architecture), it is better to involve another reviewer (I'm afraid my technical difficulties will not be fixed in the near future).

Wouldn't we also like to rename the application to match the package name?

Do not forget to update the readme later (both screenshots and functionality description). :)

app/src/main/java/com/github/kpdandroid/ui/BottomMenu.kt
TimPushkin: I noticed that Google tend to remove `is` for booleans in Compose, so doing the same in our project. Not really sure about this.

app/src/main/java/com/github/kpdandroid/MainActivity.kt
TimPushkin: I don't think that overrides need to be documented in general. Expecially overrides of Android built-in methods like this.

And yes. this is more about other developers who might wonder why is this used instead of the usual `onResume`.

app/src/main/java/com/github/kpdandroid/ui/DetectionResultColumn.kt
TimPushkin: I use `Pair` in `ImageAnalysisViewModel` because it interracts with exactly two layers and distinguishes them based on their position in the pair: this logic, if replaced with a collection, will either be broken, or become unsafe. On the contrary, these composables can work with arbitrary amount of layers and don't need to know what each layer represents.

app/src/main/java/com/github/kpdandroid/ui/viewmodels/FileAnalysisViewModel.kt
TimPushkin: 1. I think, it's obvious that zero represents no progress. And as no logic in the app is based on this value (Compose handles this for us, it'll even coerse the wrong value, if any), it is another point for leaving it as it is. So, I think, adding a constant will only make this look more complicated.
2. `Float` is the easiest type to handle here as Compose progress bars require a `Float` in range [0, 1].
3. I don't get you, what `Share` would mean here?

app/src/main/java/com/github/kpdandroid/ui/screens/FileAnalysisScreen.kt
TimPushkin: This is actually how such variables are named in [Google's examples](https://developer.android.com/reference/kotlin/androidx/compose/material/package-summary#AlertDialog(kotlin.Function0,kotlin.Function0,androidx.compose.ui.Modifier,kotlin.Function0,kotlin.Function0,kotlin.Function0,androidx.compose.ui.graphics.Shape,androidx.compose.ui.graphics.Color,androidx.compose.ui.graphics.Color,androidx.compose.ui.window.DialogProperties))

app/src/main/java/com/github/kpdandroid/ui/screens/FileAnalysisScreen.kt
TimPushkin: Maybe we can extract `5` here to a constant, but yeah, seems like it won't make anything simpler. As no other code depends on this value, I think we'll leave it as it is.

app/src/main/java/com/github/kpdandroid/ui/screens/FileAnalysisScreen.kt
TimPushkin: I'll rename it to `opened` by analogy with `expanded` in Compose's `ExposedDropdownMenu`, for example

app/src/main/java/com/github/kpdandroid/ui/screens/FileAnalysisScreen.kt
TimPushkin: Will rename to `visible`

TimPushkin:
About `README`, I've already created an issue about this a week ago: #53. Probably will fix this before creating a PR to main.

app/src/main/java/com/github/kpdandroid/ui/viewmodels/FileAnalysisViewModel.kt
alexdtat: > 1. I think, it's obvious that zero represents no progress. And as no logic in the app is based on this value (Compose handles this for us, it'll even coerse the wrong value, if any), it is another point for leaving it as it is. So, I think, adding a constant will only make this look more complicated.
> 2. `Float` is the easiest type to handle here as Compose progress bars require a `Float` in range [0, 1].
> 3. I don't get you, what `Share` would mean here?

1. Ok
2. Ok
3. I mean that this value is not used for the number of frames processed, not for the time, etc., but as the value of the fraction `done / total`. But if it's a common name, then no problem.

app/src/main/java/com/github/kpdandroid/ui/BottomMenu.kt
alexdtat: > I noticed that Google tend to remove `is` for booleans in Compose, so doing the same in our project. Not really sure about this.

Ok, let's trust Google. 👍

app/src/main/java/com/github/kpdandroid/ui/viewmodels/FileAnalysisViewModel.kt
TimPushkin: Yeah, such thing is named "progress" everywhere in the UI progress, so I think it's ok. Seems pretty understandable for me as it's a float.

alexdtat (approved):
Everything looks great!

PR #69, opened 2022-08-24
Alpha release

TimPushkin:
Implements keypoint detection app with the following capabilities:
- Keypoint detection on image files with logging
- Keypoint detection on camera snapshots in soft-realtime


app/src/main/java/com/github/kpdandroid/MainActivity.kt
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        cameraHandler.tieCameraLifecycleIfNeededTo(
+            owner.apply {
+                lifecycle.addObserver(
+                    object : LifecycleEventObserver {
+                        override fun onStateChanged(owner: LifecycleOwner, event: Lifecycle.Event) {
+                            if (event == Lifecycle.Event.ON_DESTROY) {
+                                Log.i(TAG, "Untying camera lifecycle from $owner.")
+                                cameraHandler.untieCameraLifecycleIfNeededFrom(owner)
```
DmiitriyJarosh: Provide brief explanation about this part (about 'tie') and also add nested level check in detekt - too much here))

app/src/main/java/com/github/kpdandroid/MainActivity.kt
```diff
@@ -0,0 +1,157 @@
...
+            TAG,
+            "Trying to start camera analysis (permissions granted: $isCameraPermissionGranted)."
+        )
+
+        when {
+            isCameraPermissionGranted -> cameraHandler.startImageAnalysis(
+                targetResolution = prefs.resolution,
+                callback = prefs::resolution::set
+            )
+            shouldRationalize(Manifest.permission.CAMERA) -> Toast.makeText(
```
DmiitriyJarosh, line 130: in production apps people usually add bottomsheet here with link to settings for this app for fast permission setup

app/src/main/java/com/github/kpdandroid/MainActivity.kt
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    private fun isPermissionGranted(permission: String) =
+        ContextCompat.checkSelfPermission(this, permission) == PackageManager.PERMISSION_GRANTED
+
+    private fun shouldRationalize(permission: String) =
+        (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.M) &&
+            shouldShowRequestPermissionRationale(permission)
+
+    // Not using onResume to account for changes made in split screen
```
DmiitriyJarosh, line 146: amazing! support for split screen is rare case, you are great

app/src/main/java/com/github/kpdandroid/ui/KeypointPainter.kt
```diff
@@ -0,0 +1,45 @@
...
+    fun updateImage(image: ImageBitmap?) {
+        image?.asAndroidBitmap().let { bitmap ->
+            canvas.nativeCanvas.setBitmap(bitmap)
+            this.bitmap = bitmap
+        }
+    }
+
+    fun draw(points: List<Offset>) {
+        bitmap?.eraseColor(NativeColor.TRANSPARENT)
+        canvas.drawPoints(
```
DmiitriyJarosh, line 39: it will be ideal to use opengl shader for point painting over frames... but it is just a dream

app/src/main/java/com/github/kpdandroid/ui/KeypointPainter.kt
```diff
@@ -0,0 +1,45 @@
...
+import androidx.compose.ui.graphics.asAndroidBitmap
+import androidx.compose.ui.graphics.nativeCanvas
+import android.graphics.Canvas as NativeCanvas
+import android.graphics.Color as NativeColor
+
+private const val KEYPOINT_WIDTH = 10f
+
+class KeypointPainter(image: ImageBitmap? = null, pointColor: Color = Color.Blue) {
+    private var bitmap = image?.asAndroidBitmap()
+    private val canvas = image?.let { Canvas(it) } ?: Canvas(NativeCanvas())
```
DmiitriyJarosh, line 18: why do you use compose canvas instead native one?

app/src/main/java/com/github/kpdandroid/ui/screens/NavScreen.kt
```diff
@@ -0,0 +1,66 @@
...
+@Composable
+fun NavScreen(onDestinationSelected: (NavDestination) -> Unit) {
+    Scaffold(
+        topBar = {
+            TopAppBar(
+                title = { Text(stringResource(R.string.app_name)) }
+            )
+        }
+    ) { paddingValues ->
+        LazyColumn(modifier = Modifier.padding(paddingValues)) {
```
DmiitriyJarosh, line 44: Why do you use lazy column here?

lib/src/main/java/com/github/kpdlib/traditional/Surf.kt
```diff
@@ -0,0 +1,13 @@
...
+import com.github.kpdlib.KeypointDetector
+
+/**
+ * Wrapper for a native
+ * [OpenCV SURF detector](https://docs.opencv.org/4.6.0/d5/df7/classcv_1_1xfeatures2d_1_1SURF.html).
+ *
+ * - Keypoints have sizes and angles
+ * - Descriptor size is 64
+ */
+class Surf(width: Int, height: Int) :
```
DmiitriyJarosh, line 12: Can all algos be sealed classes?

app/src/main/java/com/github/kpdandroid/MainActivity.kt
TimPushkin: CameraX uses a `LifecycleOwner` to automatically open and close cameras according to its lifecycle events. So here I tie cameras to the lifecycle of a `BackStackEntry` which corresponds to the camera analysis screen.

app/src/main/java/com/github/kpdandroid/MainActivity.kt
TimPushkin: Added an "Open settings" button that opens app's section in Settings (haven't actually find an easy way to open permissions themselves or highlight them in the Settings screen):
image

app/src/main/java/com/github/kpdandroid/ui/KeypointPainter.kt
TimPushkin: Native canvas' `drawPoints` method is much less convenient to use: it accepts an array of floats which represent 2D coordinates but aren't  even paired, whereas the canvas from Compose accepts a `List<Offset>` which is very natural for such data

app/src/main/java/com/github/kpdandroid/ui/screens/NavScreen.kt
TimPushkin: Well, since we are currently only showing two items it's pretty useless, I agree, but I want to leave this as is in case we add more screens later

lib/src/main/java/com/github/kpdlib/traditional/Surf.kt
TimPushkin: Yes, they can, but I don't see why they should be. I don't see anything in the library that will break if a user would subclass an algorithm, so why restrict them to do so?

TimPushkin:
Seems like we won't merge this

